PR #178, opened 2021-01-24
Fix GenerateMetadataSource.cmd 

dotMorten:
Not sure why this says `pwsh.exe`. That didn't work on my PC, so wonder if it's a local tool you have? The other scripts says `powershell.exe` and it works for me after changing it.


dotMorten:
Closing. Just realizing this is powershell core, and the old powershell does fail after a while. I should read the instructions closer :-)

PR #194, opened 2021-01-27
Ensure msbuild import applies to net35 projects

AArnott:
Contributes to fixing microsoft/CsWin32#68

Also use `buildTransitive/` so that installation into a projection that then chains it into a customer project will retain the msbuild imports from this project.


AArnott:
I have fully tested this locally.

kennykerr (approved):

PR #200, opened 2021-01-30
Adding empty Directory.Build.props/targets to avoid picking up external ones

tannergooding:
Without these, building the repo can pick up a `Directory.Build.props/targets` that exist in a parent directory and cause downstream build failures.

This is going to be necessary for me to resolve: https://github.com/microsoft/ClangSharp/issues/195

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project>
+
+  <!--
+    Directory.Build.props is automatically picked up and imported by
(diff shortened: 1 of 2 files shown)


tannergooding:
ClangSharp side is https://github.com/microsoft/ClangSharp/pull/198, at which point it should be possible for `win32metadata` to pull latest sources rather than having binaries checked in

tannergooding:
It's also worth noting that this repo has a `.gitmodules` file but doesn't checkin the `ext/sdk-api` modules file nor does it give explicit instructions that said submodule needs to be added for things to function.

AArnott (approved):

PR #319, opened 2021-03-09
Add metadata for registry APIs

Robo210:
Add metadata about registry value types, create/open flags, security flags, and key disposition.


Robo210:
Tangential question: is there any particular reason why the registry APIs are under Windows.Win32.WindowsProgramming as opposed to say Windows.Win32.Registry, similar to how the file APIs were put under Windows.Win32.FileSystem?

Robo210:
For what it's worth, it doesn't seem that the constant scraper from #327 actually caughts any of these enums other than REG_OPTION*, and even then it missed several values.

sotteson1:
> Tangential question: is there any particular reason why the registry APIs are under Windows.Win32.WindowsProgramming as opposed to say Windows.Win32.Registry, similar to how the file APIs were put under Windows.Win32.FileSystem?

The original organization was based on scraping the docs markup files which are used for the online help. The registry stuff comes in a section named "winprog". That said, if it makes sense to move things into a new namespace, we are fine doing that. The docs got me a long ways to getting things into namespaces but it's not perfect.

sotteson1:
> For what it's worth, it doesn't seem that the constant scraper from #327 actually caughts any of these enums other than REG_OPTION*, and even then it missed several values.

I see that the scanner did catch most of these, although some of them ended up in another enum so it can look like it missed them. Your changes look good but they need to get re-based on the current master. Then I would like to clone it and try out your changes.

sotteson1 (approved):

sotteson1:
Thanks @Robo210! This is great addition to the metadata.

PR #414, opened 2021-04-12
Fix #66 generate metadata for external packages starting with Win32 WebView2

### LINKED ISSUE #66: Supporting "Win32" APIs not included with Windows SDK
How do we generate a `.winmd` for Win32-style components that don't ship with the Windows SDK and thus won't be included in the `windows.win32.winmd` that this project produces? Here's an example of such an API:

https://www.nuget.org/packages/Microsoft.Web.WebView2

Hopefully most such components will provide WinRT APIs, but it sounds like Project Reunion may also eventually provide Win32-style APIs, so we may need a story for this. 

Such components would typically also refer to common Win32 types so we'd want them to refer to `windows.win32.winmd` rather than duplicate those definitions.

wravery:
Fix #66 

Running `.\DoAll.ps1` should generate a second metadata file in `.\bin\Microsoft.Web.WebView2.winmd` which references the types and methods from `.\bin\Windows.Win32.winmd`. Additional metadata files can be generated by adding a partition under `.\external\Partitions` with a name matching the namespace and optionally the NuGet package name. If building from a NuGet package, add a reference to `.\external\scraper\packages.config` and it will be automatically restored.

The `.\scripts\GenerateMetadataSource.ps1` and `.\scripts\BuildMetadataBin.ps1` scripts each have a new command line switch `-externalOnly`, so if you have already run the big `Windows.Win32.winmd` build and you just want to iterate on an external partition you can call those two repeatedly with `-externalOnly`.

I updated the `launchSettings.json` for `.\sources\ClangSharpSourceToWinmd\` to run the WebView2 external build under a new `BuildExternal` configuration.

I needed to workaround https://github.com/dotnet/roslyn/issues/52516 for now, so I declared an `autoType` that maps `tagVARIANT` to `Windows.Win32.Automation.DUMMY_VARIANT` instead of `Windows.Win32.Automation.VARIANT` for external partitions. Once that bug is fixed and the `ClangSharpSourceToWinmd` references are updated with a Roslyn build that has the fix, that remap should be removed.


wravery:
Fixes #66

kennykerr:
This looks interesting. I would just point out that to really fix #66 I was hoping for a standalone tool e.g. "midl32" that would work much like midlrt does today by parsing IDL and producing a winmd directly. We would ideally provide a simple tool that component authors can easily include in their build pipelines to spit out a winmd to add to their existing published packages.

wravery:
> This looks interesting. I would just point out that to really fix #66 I was hoping for a standalone tool e.g. "midl32" that would work much like midlrt does today by parsing IDL and producing a winmd directly. We would ideally provide a simple tool that component authors can easily include in their build pipelines to spit out a winmd to add to their existing published packages.

Ah, I see. To get the references to `Windows.Win32.winmd` to work though it would still be coupled to this project.

I wonder if there's a way to do this with [Type Library](https://docs.microsoft.com/en-us/previous-versions/windows/desktop/automat/contents-of-a-type-library) files built from MIDL v2 instead of going directly from the IDL. They're sort of the equivalent to Windows Metadata files for COM/MIDL v2.

kennykerr:
> Ah, I see. To get the references to Windows.Win32.winmd to work though it would still be coupled to this project.

Yes, the winmd file is the source of truth and the point of interchange between projects. This is the pattern WinRT established. How those winmd files are produced can vary dramatically from project to project, but the dependency should be limited to consuming the winmd and not necessarily the tools used to build that winmd.

wravery:
> How those winmd files are produced can vary dramatically from project to project, but the dependency should be limited to consuming the winmd and not necessarily the tools used to build that winmd.

I don't think that's possible, though. If everything were in a single namespace, then you might just need to provide that namespace to an external tool and it could reference the types from there. But as long as everything is split into sub-namespaces/tech-roots (a good thing, not suggesting they shouldn't be) and there's any remapping or autotypes, references to those types depend on the tools that generated the winmd.

kennykerr:
Yes, this will require more formality around how Win32 metadata is defined. This is unambiguous for WinRT and has worked that way for years, but it will require more spec'ing for Win32 to succeed in this same way.

wravery:
Perhaps this project could produce a standalone tool which carries the mapping information, either as a config file or statically embedded.

wravery:
> I wonder if there's a way to do this with [Type Library](https://docs.microsoft.com/en-us/previous-versions/windows/desktop/automat/contents-of-a-type-library) files built from MIDL v2 instead of going directly from the IDL. They're sort of the equivalent to Windows Metadata files for COM/MIDL v2.

I looked into this some more, and I don't think it would work. TLB files don't seem to include any free functions (vs. methods) or constants. Working directly from the IDL also seems like a non-starter because there's a lot of information encoded in `cpp_quote` directives which only shows up once the IDL has been run through `midl.exe` to produce the C++ headers. I think ultimately the same `ClangSharp` approach needs to be re-used for external packages/headers to get the best results.

wravery:
Not sure how yet, but something in this PR has triggered the bug I reported as https://github.com/dotnet/roslyn/issues/52516. If I checkout master and rebuild from a clean repo (or use a Windows.Win32.winmd from one of the published NuGet packages), that bug no longer repros. To be safe, I'm closing this PR, I'll reopen it if/when I figure out what changed in the base Windows.Win32.winmd file.

PR #426, opened 2021-04-17
Add ManagementInfrastructure header

wmgries:
Add the [Management Infrastructure](https://docs.microsoft.com/en-us/windows/win32/api/_wmi_v2/) API header (mi.h). Although Management Infrastructure API provides similar/the same functionality as WMI, PowerShell uses the Management Infrastructure through through the [MMI](https://github.com/powershell/mmi) project rather than the WMI APIs. I'm not an expert on the rationale, but it appears the MI APIs are positioned as a "WMI v2"; this [blog post](https://devblogs.microsoft.com/scripting/should-i-use-cim-or-wmi-with-windows-powershell/) appears to explain some of the differences.

I picked "ManagementInfrastructure" to match the MMI namespace (Microsoft.Management.Infrastructure). Many of the structs/functions/etc. were already in the namespace mapping file being mapped to Windows.Win32.Wmi (although not included in the build, because mi.h wasn't referenced), so I changed their mapping. I don't have any strong view on this naming, but this made sense to me.


sotteson1 (approved):

sotteson1:
Thanks @wmgries!

PR #450, opened 2021-04-28
generation: Add DirectXShaderCompiler partition

MarijnS95:
dxcapi.h is shipped with the SDK and used to interact with `dxcompiler.dll`: https://github.com/microsoft/DirectXShaderCompiler.

### DIFF
--- a/generation/scraper/Partitions/DirectXShaderCompiler/main.cpp
+++ b/generation/scraper/Partitions/DirectXShaderCompiler/main.cpp
@@ -1,0 +1,10 @@
+#define SECURITY_WIN32 // For sspi.h
+#define QCC_OS_GROUP_WINDOWS
+
+#include "intrinfix.h"
+
+#include "windows.fixed.h"
+#include <sdkddkver.h>
+
+#include <dxcapi.h>
+// #include <dxcisense.h>
--- a/generation/scraper/Partitions/DirectXShaderCompiler/settings.rsp
+++ b/generation/scraper/Partitions/DirectXShaderCompiler/settings.rsp
@@ -1,0 +1,11 @@
+--file
+<PartitionDir>\main.cpp
+--output
+<GeneratedSourceDir>\<PartitionName>.cs
+--include-directory
+<RepoRoot>\generation\scraper
+<IncludeRoot>/um
+--traverse
+<IncludeRoot>/um/dxcapi.h
+--namespace
+Windows.Win32.Graphics.Hlsl


generation/scraper/Partitions/DirectXShaderCompiler/main.cpp
```diff
@@ -0,0 +1,10 @@
+#define SECURITY_WIN32 // For sspi.h
+#define QCC_OS_GROUP_WINDOWS
+
+#include "intrinfix.h"
+
+#include "windows.fixed.h"
+#include <sdkddkver.h>
+
+#include <dxcapi.h>
+// #include <dxcisense.h>
```
MarijnS95, line 10: Unfortunately the IntelliSense headers aren't bundled (yet?) but those woudl be needed to fully convert [hassle-rs](https://github.com/Traverse-Research/hassle-rs/) over to `windows-rs`.

Perhaps those can be included separately when it's easier to generate [loose metadata files](https://github.com/microsoft/windows-rs/tree/master/.windows/winmd) following #414.

generation/scraper/Partitions/DirectXShaderCompiler/settings.rsp
```diff
@@ -0,0 +1,11 @@
...
+<PartitionDir>\main.cpp
+--output
+<GeneratedSourceDir>\<PartitionName>.cs
+--include-directory
+<RepoRoot>\generation\scraper
+<IncludeRoot>/um
+--traverse
+<IncludeRoot>/um/dxcapi.h
+--namespace
+Microsoft.DirectXShaderCompiler
```
MarijnS95, line 11: Is this the right namespace? I'm not sure :)

sotteson1:
We already have a partition named Windows.Win32.Graphics.Hlsl that includes d3dcompiler.h. Should this file go in there instead?

MarijnS95:
@sotteson1 That might work, those types are prefixed with `D3D` whereas this header exclusively contains `(I)Dxc`-prefixed types. Not sure if it's cleaner to put it in a separate namespace like `Windows.Win32.Graphics.Dxc`, but then it deals exclusively with HLSL inputs too. Split these out in `Windows.Win32.Graphics.Hlsl.D3D` and `Windows.Win32.Graphics.Hlsl.Dxc`?

mikebattista:
My recommendation would be to put it in the existing Hlsl namespace.

generation/scraper/Partitions/DirectXShaderCompiler/settings.rsp
mikebattista: Please move this to Windows.Win32.Graphics.Hlsl.

MarijnS95: @mikebattista Updated to that, and regenerated the baseline too.

MarijnS95:
@mikebattista did I make a mistake during regeneration or is the file polluted by the environment it was generated in?

mikebattista:
I just merged in the PR. There were a few new DXC constants missing from the baseline when I built, so I just added them. Everything else looked good.

MarijnS95:
@mikebattista just curious what was missing/changed since the file is binary. It decreased in size as well, how does that equate to more constants being added? Did I build from the wrong Windows headers (since I did not check out the `ext/sdk-api` submodule locally)?

Would like to know in case of future contributions :)

mikebattista:
Below were the missing values. I just ran `./DoAll.ps1 -Clean` which reported the below deltas.

DXC_CP_UTF8 not found in 1st winmd
DXC_CP_UTF16 not found in 1st winmd
DXC_CP_ACP not found in 1st winmd
DXC_HASHFLAG_INCLUDES_SOURCE not found in 1st winmd

MarijnS95:
@mikebattista ah, I just spotted those too trying to work around crashing AvaloniaILSpy on Linux :)

Thanks! I actually need these flags. Still curious why there was a size decrease, did some other useless garbage get omitted again? Will probably double-check with `ext/sdk-api` next time around :)

MarijnS95:
Oh, by the way, since I was working with these flags, should we add an annotation specifying that these are used in `CreateBlobWithEncodingFromPinned` and returned by `IDxcBlobEncoding::GetEncoding`?

mikebattista:
I'm not sure about the size decrease. I just followed the usual steps for updating the baseline.

@sotteson1 thoughts on this and the annotations? If there are annotations needed, what's the recommendation there?

MarijnS95:
Should be trivial to replace them with enums as documented under https://github.com/microsoft/win32metadata#how-you-can-help. I'll try it out when booted into Windows tomorrow, something like:

```cs
public enum DXC_CODEPAGE
{
    DXC_CP_UTF8 = DXC_CP_UTF8,
    DXC_CP_UTF16 = DXC_CP_UTF16,
    DXC_CP_ACP = DXC_CP_ACP,
}
```

Will have to check how to namespace the constants so that they are reused in the enum.

By the way, DXC simply redefines the `CP_` constants from Windows headers. When doing this we should probably define it in a generic place and use it for `SetConsoleOutputCP`/`GetConsoleOutputCP` (if at all generated) and perhaps other functions too.

mikebattista:
Our enum support did change recently so that documentation you referenced is a bit out-of-date. @sotteson is working on updating that which we'll add to our new [CONTRIBUTING.md](https://github.com/microsoft/win32metadata/blob/master/CONTRIBUTING.md).

Let's agree on the right design with @sotteson then we can make the necessary changes.

MarijnS95:
@mikebattista regarding missing constants, they are not in the `artifacts\InstalledPackages\Microsoft.Windows.SDK.CPP.10.0.19041.5\c\Include\10.0.19041.0\um\dxcapi.h` header. Should I generate the baseline from a different SDK, too? That might also explain the size difference. It looks like the constants appeared 2 years ago in https://github.com/microsoft/DirectXShaderCompiler/commit/f4965b71dd878c42c219ec00d31c406c483fd045#diff-464695ecc4ab8caa605f8837a218e297f4450d4d59a9f2cd0675f9a1b0362b75R85-R88, that's why they only showed up recently. Not that we need these anyway, they are redefining constants from `Windows.Win32.Globalization.Apis` and I'd rather leverage those in a more generic way.

MarijnS95:
@mikebattista @sotteson1 Figured it out in the end, only needed to add it to `enums.json`: https://github.com/MarijnS95/win32metadata/compare/codepage.

However:
- This puts the `CODEPAGE` enum in the `Hlsl` namespace - any way to move it to `Globalization`? Preferably without defining it by hand in `Globalization.Manual.cs` and using `"addUsesTo`";
- Should we make a separate enum for DXC after all? Newer `dxcapi.h` headers only redefine `DXC_CP_UTF7/8/ACP`, and perhaps throw an error on other types - but I have not validated that.

mikebattista:
> 
> Two more questions:
> 
> * Perhaps DXC needs its own enum after all, to signify that only `UTF8`, `UTF16` codepages and `ACP` may be passed? (I haven't checked the source to back this up);
> * Are there plans to propagate annotations like these back into the headers the metadata was generated from?
>   DXC is open source so we might as well submit a PR there to get this (and other flags) wrapped in enums if that makes things easier. The other two are `DxcValidatorFlags_*` and `DxcVersionInfoFlags_*`.

If DXC defined its own constants with a DXC prefix, I'd prefer to keep that design and keep the enum specific to DXC. Ideally yes the C APIs used actual enums instead of just #defines. We can't really break existing headers in the Windows SDK, if DXC is more open to breaking changes, then sure feel free to pursue that.

And yes, `enums.json` is the new way to define enums. You can specify a namespace by adding a `namespace` property, otherwise it will inherit the namespace of the first entry in `uses`. I'd prefer to infer the namespace from `uses` rather than define it explicitly. With all the namespace refactoring I'm doing, it makes it easier to keep related types together.

I'll have a CONTRIBUTING.md update soon with the details.

MarijnS95:
> If DXC defined its own constants with a DXC prefix, I'd prefer to keep that design and keep the enum specific to DXC.

"Unfortunately" this is only in a newer header, we'll have to use `CODEPAGE` for now. It seems to be defined for easy access if anything.

> Ideally yes the C APIs used actual enums instead of just #defines. We can't really break existing headers in the Windows SDK

Ah right, wrapping the constants in an an `enum` shouldn't hurt but changing the types of the functions will. I think they adhere to the same standards as other COM interfaces; changes are only made on a new interface revision and this is definitely a breaking change for very little gain.

> You can specify a namespace by adding a `namespace` property

That was too obvious to miss 😬 - pretty sure I searched the file for the word `namespace` but nothing surfaced.

> otherwise it will inherit the namespace of the first entry in `uses`. I'd prefer to infer the namespace from `uses` rather than define it explicitly.

Me as well. Going with that now by providing this annotation for some functions and types from that namespace. Had to run a `-Clean` to make sure the enum doesn't show up in both `Hlsl` and `Globalization`, but the `CP_` prefix picks up some unrelated flags:

```
Duplicate constants/enum names detected:
CP_INSTALLED
  Windows.Win32.Globalization.CODEPAGE
  Windows.Win32.Globalization.ENUM_SYSTEM_CODE_PAGES_FLAGS
CP_SUPPORTED
  Windows.Win32.Globalization.CODEPAGE
  Windows.Win32.Globalization.ENUM_SYSTEM_CODE_PAGES_FLAGS
```

I think this counts as `$failed = false` because no `If all the differences are expected, please update the baseline by doing this: ...` is printed.

How do I omit those two from the `autoPopulate` `"filter": "CP_"`?

This it the result btw; someone may want to help find and annotate more types/functions in the future.

```
Windows.Win32.Graphics.Hlsl.IDxcBlobEncoding.GetEncoding : pCodePage...UInt32* => CODEPAGE*
Windows.Win32.Graphics.Hlsl.IDxcLibrary.CreateBlobFromFile : codePage...UInt32* => CODEPAGE*
Windows.Win32.Graphics.Hlsl.IDxcLibrary.CreateBlobWithEncodingFromPinned : codePage...UInt32 => CODEPAGE
Windows.Win32.Graphics.Hlsl.IDxcLibrary.CreateBlobWithEncodingOnHeapCopy : codePage...UInt32 => CODEPAGE
Windows.Win32.Graphics.Hlsl.IDxcLibrary.CreateBlobWithEncodingOnMalloc : codePage...UInt32 => CODEPAGE
Windows.Win32.Globalization.CPINFOEXA.CodePage...UInt32 => CODEPAGE
Windows.Win32.Globalization.CPINFOEXW.CodePage...UInt32 => CODEPAGE
Windows.Win32.Globalization.CODEPAGE not found in 1st winmd
Windows.Win32.Globalization.Apis.GetCPInfo : CodePage...UInt32 => CODEPAGE
Windows.Win32.Globalization.Apis.GetCPInfoExA : CodePage...UInt32 => CODEPAGE
Windows.Win32.Globalization.Apis.GetCPInfoExW : CodePage...UInt32 => CODEPAGE
```

PR #474, opened 2021-05-13
scraper/enums: Define enum for Windows.Win32.Graphics.Hlsl.DXC_CP

MarijnS95:
From the discussion in https://github.com/microsoft/win32metadata/pull/450#issuecomment-831501666

This results in the following enum:
```cs
// Windows.Win32.Graphics.Hlsl.DXC_CP
public enum DXC_CP : uint
{
	DXC_CP_UTF8 = 65001u,
	DXC_CP_UTF16 = 1200u,
	DXC_CP_ACP = 0u
}
```

With the following replacements:

```
Windows.Win32.Graphics.Hlsl.IDxcBlobEncoding.GetEncoding : pCodePage...UInt32* => DXC_CP*
Windows.Win32.Graphics.Hlsl.IDxcLibrary.CreateBlobFromFile : codePage...UInt32* => DXC_CP*
Windows.Win32.Graphics.Hlsl.IDxcLibrary.CreateBlobWithEncodingFromPinned : codePage...UInt32 => DXC_CP
Windows.Win32.Graphics.Hlsl.IDxcLibrary.CreateBlobWithEncodingOnHeapCopy : codePage...UInt32 => DXC_CP
Windows.Win32.Graphics.Hlsl.IDxcLibrary.CreateBlobWithEncodingOnMalloc : codePage...UInt32 => DXC_CP
Windows.Win32.Graphics.Hlsl.DXC_CP not found in 1st winmd
```

<details>
<summary>Old comment when this was done for all of `CP_`</summary>

This results in the following enum:

```cs
// Windows.Win32.Globalization.CODEPAGE
public enum CODEPAGE : uint
{
    CP_INSTALLED = 1u,
    CP_SUPPORTED = 2u,
    CP_ACP = 0u,
    CP_OEMCP = 1u,
    CP_MACCP = 2u,
    CP_THREAD_ACP = 3u,
    CP_SYMBOL = 42u,
    CP_UTF7 = 65000u,
    CP_UTF8 = 65001u
}
```

With the following replacements:

    Windows.Win32.Graphics.Hlsl.IDxcBlobEncoding.GetEncoding : pCodePage...UInt32* => CODEPAGE*
    Windows.Win32.Graphics.Hlsl.IDxcLibrary.CreateBlobFromFile : codePage...UInt32* => CODEPAGE*
    Windows.Win32.Graphics.Hlsl.IDxcLibrary.CreateBlobWithEncodingFromPinned : codePage...UInt32 => CODEPAGE
    Windows.Win32.Graphics.Hlsl.IDxcLibrary.CreateBlobWithEncodingOnHeapCopy : codePage...UInt32 => CODEPAGE
    Windows.Win32.Graphics.Hlsl.IDxcLibrary.CreateBlobWithEncodingOnMalloc : codePage...UInt32 => CODEPAGE
    Windows.Win32.Globalization.CPINFOEXA.CodePage...UInt32 => CODEPAGE
    Windows.Win32.Globalization.CPINFOEXW.CodePage...UInt32 => CODEPAGE
    Windows.Win32.Globalization.CODEPAGE not found in 1st winmd
    Windows.Win32.Globalization.Apis.GetCPInfo : CodePage...UInt32 => CODEPAGE
    Windows.Win32.Globalization.Apis.GetCPInfoExA : CodePage...UInt32 => CODEPAGE
    Windows.Win32.Globalization.Apis.GetCPInfoExW : CodePage...UInt32 => CODEPAGE

---

As mentioned in https://github.com/microsoft/win32metadata/pull/450#issuecomment-840410662 `"filter": "CP_"` picks up two unrelated constants already defined in another enum:

```
Duplicate constants/enum names detected:
CP_INSTALLED
  Windows.Win32.Globalization.CODEPAGE
  Windows.Win32.Globalization.ENUM_SYSTEM_CODE_PAGES_FLAGS
CP_SUPPORTED
  Windows.Win32.Globalization.CODEPAGE
  Windows.Win32.Globalization.ENUM_SYSTEM_CODE_PAGES_FLAGS
```

### TODO
- Is the name `CODEPAGE` okay?
- Any more functions/types that need to be annotated?

</details>

### Note on DXC

- Keep in mind newer DXC headers redefines three of these constants: https://github.com/microsoft/DirectXShaderCompiler/blob/aa4f5176ebba010a860a214d73f33bdc5b5042ce/include/dxc/dxcapi.h#L90-L94
  These are the only supported codepages, others are invalid.
- A new [`IDxcUtils` interface also takes in codepage](https://github.com/microsoft/DirectXShaderCompiler/blob/aa4f5176ebba010a860a214d73f33bdc5b5042ce/include/dxc/dxcapi.h#L363-L390) arguments. These need to be annotated too, can that already be done in advance or will the emitter complain on unmatched methods in the `"uses"` list?


mikebattista:
New contributing guidelines are up at https://github.com/microsoft/win32metadata/blob/master/CONTRIBUTING.md.

I don't think the CP_INSTALLED and CP_SUPPORTED should be in the same enum as the others. As the duplicate shows, it's used by the enum API.

Also the GetCPInfo APIs take an open set of values it seems from the docs, not just the ones from the enum. @sotteson1 is an enum the proper choice here for an open set?

MarijnS95:
> New contributing guidelines are up at https://github.com/microsoft/win32metadata/blob/master/CONTRIBUTING.md.

That's nice and useful, thanks!

> I don't think the CP_INSTALLED and CP_SUPPORTED should be in the same enum as the others. As the duplicate shows, it's used by the enum API.

Correct, these are explicitly listed in `enums.json` under `ENUM_SYSTEM_CODE_PAGES_FLAGS`. Asked above if there's an easy way to omit them from `autoPopulate`'s `filter`?

> Also the GetCPInfo APIs take an open set of values it seems from the docs, not just the ones from the enum. @sotteson1 is an enum the proper choice here for an open set?

You're right, [the docs](https://docs.microsoft.com/en-us/windows/win32/intl/code-page-identifiers) mention a lot more values than are available in constants, perhaps it's not a good idea to define this for `CODEPAGE` then. OTOH DXC explicitly supports three pages and nothing more, so we could have a `DXC_CP` enum here then.

The values are not in the SDK header yet but should hopefully be in some future release, does it make sense to list the members explicitly now and have `"filter": "DXC_CP_"` already in place, so that it hopefully errors out on a conflict when the manual members are ready to be removed again?

Should I order the members by value, for cleanliness? Copypasted them from [the header](https://github.com/microsoft/DirectXShaderCompiler/blob/aa4f5176ebba010a860a214d73f33bdc5b5042ce/include/dxc/dxcapi.h#L90-L94) but that seems to be arbitrary ordering anyway.

mikebattista:
> You're right, [the docs](https://docs.microsoft.com/en-us/windows/win32/intl/code-page-identifiers) mention a lot more values than are available in constants, perhaps it's not a good idea to define this for `CODEPAGE` then. OTOH DXC explicitly supports three pages and nothing more, so we could have a `DXC_CP` enum here then.

I prefer the DXC_CP approach with constrained values just for DXC.

> The values are not in the SDK header yet but should hopefully be in some future release, does it make sense to list the members explicitly now and have `"filter": "DXC_CP_"` already in place, so that it hopefully errors out on a conflict when the manual members are ready to be removed again?

I would just manually define them for now. We can make future changes to the metadata when the time comes if necessary.

> Should I order the members by value, for cleanliness? Copypasted them from [the header](https://github.com/microsoft/DirectXShaderCompiler/blob/aa4f5176ebba010a860a214d73f33bdc5b5042ce/include/dxc/dxcapi.h#L90-L94) but that seems to be arbitrary ordering anyway.

I would order them by value (0 first) so the enum members are ordered by value as you'd expect.

MarijnS95:
> I would just manually define them for now. We can make future changes to the metadata when the time comes if necessary.

They're manually defined now, but I feel like keeping `"filter": "DXC_CP_"` so that this extra bit of automation is not forgotten about when the next SDK (or perhaps the one after) includes these.

> I would order them by value (0 first) so the enum members are ordered by value as you'd expect.

👍 They're in the right order now.

> I prefer the DXC_CP approach with constrained values just for DXC.

Cool, then we're on the same page, and this PR is ready for review and merging :)

mikebattista:
Merged. Thanks for the contribution.

teoxoy:
@MarijnS95 I think `DxcBuffer.Encoding` should also be `DXC_CP` instead of `u32` but I'm not 100% sure. Thoughts?

MarijnS95:
@teoxoy for completeness, beb18b9d045371cb1374a721be5f9d50fcdd4f45 added `DxcBuffer` exactly 5 months _after_ this PR was opened :)

I think you are right, at least the `// Use Encoding = 0 for non-text bytes, ANSI text, or unknown with BOM.
` description for ANSI and UTF BOM autodetection matches the comment on `DXC_CP_ACP`.

More DXC annotations appear to be missing; I cobbled together some in https://github.com/microsoft/win32metadata/pull/1989 but without any codegen/testing for now.

PR #476, opened 2021-05-14
Use nerdbank.gitversioning to better deal with versioning

sotteson1:
Thanks to @AArnott for this cool project that automatically increments our build numbers for us.


sources/msbuild/version.json
```diff
@@ -0,0 +1,11 @@
+{
+  "$schema": "https://raw.githubusercontent.com/dotnet/Nerdbank.GitVersioning/master/src/NerdBank.GitVersioning/version.schema.json",
+  "version": "0.2-preview",
```
AArnott, line 3: As this is only different from your repo-root file by the versions, you cloud remove all the redundant properties that follow this line and replace with `"inherit": true` if you wanted.

PR #494, opened 2021-05-26
Add missing MI_FLAG_* constants.

wmgries:
I noticed that several MI_FLAG_* constants were defined in the mi.h header file but not making into the generated winmd file, namely:

```C++
#define MI_FLAG_ANY             (1|2|4|8|16|32|64)
#define MI_FLAG_VERSION         (1<<26|1<<27|1<<28)
#define MI_FLAG_ADOPT	        ((MI_Uint32)(1 << 31))
```

I suspect their use of bitwise or is what's causing them to be excluded. I did not do a thorough scan of the header to see if other constants are missing. Definitions were directly copied from the header file into the manual C# file.


wmgries:
@mikebattista @sotteson1 if this looks good, can you merge? It takes a bit to keep up with the winmd updates since this is a big binary file :)

PR #497, opened 2021-05-28
Log the ClangSharpPInvokeGenerator path and ensure the script doesn't fail for missing tools

tannergooding:
These were found while testing https://github.com/microsoft/ClangSharp/pull/235.

In particular, it is useful to log the resolved path for `ClangSharpPInvokeGenerator` so it can be asserted to be the expected one (ClangSharp's CI test should override the global tool, for example).

Likewise, `Select-String` does not output a `string` by default, it outputs a `Microsoft.PowerShell.Commands.MatchInfo`: <https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.utility/select-string?view=powershell-7.1#outputs> which does not have a Length property. It is expected that `-Raw` be passed if you want a `string`.

Additionally, if there is no match the result can be `$null`, which needs to be checked before getting the length.

### DIFF
--- a/scripts/CommonUtils.ps1
+++ b/scripts/CommonUtils.ps1
@@ -46,16 +46,16 @@ function Install-DotNetTool
 
     if ($Version -ne '')
     {
-        $installed = & dotnet tool list -g | select-string "$Name\s+$Version"
-        if (!$installed.Length)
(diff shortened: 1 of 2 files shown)


tannergooding:
CC. @sotteson1

sotteson1:
Thanks @tannergooding!

PR #548, opened 2021-06-23
Add API Docs package build to pipeline

AArnott:
This adds a project to build the `Microsoft.Windows.SDK.Win32Docs` nuget package to the repo and pipeline.

This produces the equivalent API docs file that the CsWin32 projection currently consumes, but instead of it being an embedded resource in CsWin32 itself, it can now be shared with all projections that wish to add API documentation to the generated projections.

Consumption may look like this:

```xml
<PackageReference Include="Microsoft.Windows.SDK.Win32Docs" Version="*" />
```

A reference to this package will also define the `MicrosoftWindowsSdkApiDocsPath` msbuild property in the consuming project, which can be consumed by the projection. This property is added as a `CompilerVisibleProperty` so that a roslyn Source Generator can observe it as well.

This package also includes as assembly that defines the `ApiDetails` class that describes the msgpack schema for each API.
The overall schema is a map where the key is a string that is the API name, and the value is the `ApiDetails` class.


azure-pipelines.yml
```diff
@@ -223,8 +232,8 @@ jobs:
       SessionTimeout: '60'
       MaxConcurrency: '50'
       MaxRetryAttempts: '2'
-      condition: eq(variables['SignFiles'], 'true')
-  
+    condition: and(succeeded(), eq(variables['SignFiles'], 'true'))
```
AArnott, line 235: The indentation was wrong on these conditions, leading to them being totally ignored.
They were also missing the `succeeded()` condition, which means that signing would have been attempted even if the pipeline had already failed.

azure-pipelines.yml
```diff
@@ -2,10 +2,15 @@ trigger:
...
-pr: none
+pr:
+- master
+
+variables:
+  BuildConfiguration: Release
 
 jobs:
 - job: scrape_x64
+  condition: false
```
AArnott, line 13: I'll revert these `condition: false` lines. They were only added to save Azure Pipelines resources while I iterated on these changes.

AArnott:
**BEFORE MERGING**: I'd like to test out CsWin32 consuming this package to verify we have everything we need here.

azure-pipelines.yml
```diff
@@ -139,8 +143,8 @@ jobs:
   - task: UseDotNet@2
     displayName: Install DotNet 2.1.x for signing tasks
     inputs:
-      packageType: 'sdk'
-      version: '2.1.x'
+      packageType: runtime
```
AArnott, line 146: `sdk` was more than required, so switching to `runtime` here presumably makes the pipeline run a little faster.

azure-pipelines.yml
```diff
@@ -334,3 +338,70 @@ jobs:
...
+    # There's a problem on microsoft.visualstudio.com that requires the guid instead of NuGetCommand@2
+    # Don't publish if we're using pre-generated source
+  - task: 333b11bd-d341-40d9-afcf-b32d5ce6f23b@2
+    displayName: 📤 NuGet push
+    inputs:
+      command: push
+      packagesToPush: $(System.DefaultWorkingDirectory)/bin/Packages/$(BuildConfiguration)/NuGet/*.nupkg
+      publishVstsFeed: c1408dcb-1833-4ae4-9af5-1a891a12cc3c
+      allowPackageConflicts: true
+    condition: and(succeeded(), ne(variables['Build.Reason'], 'PullRequest'))
```
AArnott, line 407: I'm *not* pushing packages on PR. But other jobs in this pipeline do. I would suggest as a separate PR that we avoid pushing packages anywhere on PR.
I'd also like to refactor this file to allow for publicly observable pipelines on `azure-public` so our OSS contributors can access them, the way we have for CsWin32 today.

sotteson1 (approved):

PR #549, opened 2021-06-24
Add NativeTypeName attribute to remapped structs

wravery:
I think adding the `NativeTypeName` attribute here should provide enough information to another tool to implement #66 with references back to `Windows.Win32.winmd` and `Windows.Win32.Interop.dll`.


wravery:
For example, this adds `[NativeTypeName("tagVARIANT")]` to the declaration of `struct VARIANT` in `Automation.modified.cs`. Another tool built with `ClangSharpPInvoke` would need to know about that.

wravery:
I went ahead and merged this with #553, so this one can be closed. The SDK in the `msbuild` directory should make this largely unnecessary, but it might be possible to replace some of the files exported in the SDK NuGet with data gleaned directly from the winmd in the future.

PR #553, opened 2021-06-26
Add a WebView2Sample with a StaticLibrary attribute mapping the DllImport name to an optional static library substitute

wravery:
Per https://github.com/microsoft/windows-rs/pull/729, there are times when it would be nice to link a static lib instead of importing an entry-point from the DLL. None of the libraries in `Windows.Win32.winmd` currently support this since I think they're all system DLLs, but when #66 is finished this is a capability that would make sense to include.

I tested this manually with `Windows.Win32.winmd` by mapping `USER32` to a `FakeStaticUser32` LIB in the launch settings for `ClangSharpSourceToWinmd`, then I revived the `TestSource` target and codified the test mapping `MyDll` to `MyStaticLib`.

The result is a new attribute on all of the functions imported from `MyDll` which looks like `[StaticLibrary("MyStaticLib")]`. This is in addition to the existing `[DllImport("MyDll", ...]` attributes which are unchanged, so consumers of the winmd file that don't know about `StaticLibraryAttribute` can safely ignore it.

I also merged in the change from #549 which adds a `NativeTypeName` attribute to any structs which are renamed by `ClangSharpPInvokeGenerator --remap`. (e.g. `[NativeTypeName("tagVARIANT")]` on `VARIANT`). That should reduce the dependency on files exported from this repo in the SDK NuGet, or enable an out-of-tree tool to interop with just the base winmd more easily.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -358,7 +358,7 @@ MigrationBackup/
 /sources/CsWin32ProjectionLib/ProjectedWin32.cs
 /sources/CsWin32/Properties/launchSettings.json
 /.vscode/settings.json
-/tests/SourceToWinmd/source/generated
+/tests/SourceToWinmd/source/generated/fake/autotypes.cs
 /generation/emitter/generated
 /winsdk
 /generation/scraper/autoTypes.generated.rsp
--- a/sources/ClangSharpSourceToWinmd/ClangSharpSourceCompilation.cs
+++ b/sources/ClangSharpSourceToWinmd/ClangSharpSourceCompilation.cs
@@ -35,7 +35,8 @@ namespace ClangSharpSourceToWinmd
             Dictionary<string, string> typeImports,
             Dictionary<string, string> requiredNamespaces,
             HashSet<string> reducePointerLevels,
-            IEnumerable<string> addedRefs)
+            IEnumerable<string> addedRefs,
+            Dictionary<string, string> staticLibs)
         {
             sourceDirectory = Path.GetFullPath(sourceDirectory);
 
@@ -104,7 +105,7 @@ namespace ClangSharpSourceToWinmd
                     Directory.CreateDirectory(newSubDir);
                 }
 
-                var cleanedTree = MetadataSyntaxTreeCleaner.CleanSyntaxTree(tree, remaps, enumAdditions, enumsMakeFlagsHashSet, requiredNamespaces, emptyStucts, enumMemberNames, modifiedFile);
+                var cleanedTree = MetadataSyntaxTreeCleaner.CleanSyntaxTree(tree, remaps, enumAdditions, enumsMakeFlagsHashSet, requiredNamespaces, staticLibs, emptyStucts, enumMemberNames, modifiedFile);
 
                 lock (cleanedTrees)
                 {
--- a/sources/ClangSharpSourceToWinmd/MetadataSyntaxTreeCleaner.cs
+++ b/sources/ClangSharpSourceToWinmd/MetadataSyntaxTreeCleaner.cs
@@ -11,9 +11,9 @@ namespace ClangSharpSourceToWinmd
 {
     public static class MetadataSyntaxTreeCleaner
     {
-        public static SyntaxTree CleanSyntaxTree(SyntaxTree tree, Dictionary<string, string> remaps, Dictionary<string, Dictionary<string, string>> enumAdditions, HashSet<string> enumsMakeFlags, Dictionary<string, string> requiredNamespaces, HashSet<string> nonEmptyStructs, HashSet<string> enumMemberNames, string filePath)
+        public static SyntaxTree CleanSyntaxTree(SyntaxTree tree, Dictionary<string, string> remaps, Dictionary<string, Dictionary<string, string>> enumAdditions, HashSet<string> enumsMakeFlags, Dictionary<string, string> requiredNamespaces, Dictionary<string, string> staticLibs, HashSet<string> nonEmptyStructs, HashSet<string> enumMemberNames, string filePath)
         {
-            TreeRewriter treeRewriter = new TreeRewriter(remaps, enumAdditions, enumsMakeFlags, requiredNamespaces, nonEmptyStructs, enumMemberNames);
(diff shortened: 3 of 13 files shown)


wravery:
Just found the `msbuild/GeneratorSample` sub-project, looks like that's where the fix for #66 is going. So I'll see if I can get it to generate a WebView2 winmd with the `--staticLibs` flag.

wravery:
I added a minimal sample which generates `Microsoft.Web.WebView2.Win32.winmd` using the SDK build tasks. The SDK still seems to be a WIP, but with some configuration parameters and manual builds using local binaries, I was able to build and validate the output with ILSpy.

The sample also adds the `[StaticLibrary("WebView2LoaderStatic")]` attribute from earlier commits, so it can be used with projections that support static linkage (e.g. https://github.com/microsoft/windows-rs in the future).

sotteson1:
Hey @wravery, I'm trying to understand how the static lib approach would work. How would a language projection use that? I'm thinking of C# since I've done a lot of interop for C#, and I don't know how C# could consume a static lib.

sotteson1:
> I also merged in the change from #549 which adds a `NativeTypeName` attribute to any structs which are renamed by `ClangSharpPInvokeGenerator --remap`. (e.g. `[NativeTypeName("tagVARIANT")]` on `VARIANT`). That should reduce the dependency on files exported from this repo in the SDK NuGet, or enable an out-of-tree tool to interop with just the base winmd more easily.

Why do we need this? We do a lot of renames. It seems like this is going to add a lot of bloat to the metadata.

wravery:
I don't think it would work for C# or any other managed/interpreted language projection. It's more for the Rust projection than anything else as far as I know, which does compile all the way down to native code and can do FFI against static libs as easily as DLL imports. The linker can also remove dead code and perform more optimizations. I have a prototype putting that all together against the windows-rs webview2 sample winmd in this example, it cuts down the total redistributable size quite a bit.

wravery:
> Why do we need this? We do a lot of renames. It seems like this is going to add a lot of bloat to the metadata. The whole "tag" notion is an old Win32 holdover. The "tag" versions of the struct names are nearly never used in real code.

The SDK went in a different direction than I expected by redistributing the rsp files in the NuGet. I initially made that change thinking that supporting #66 with only the winmd file would need to know about any remapping that happened in the ClangSharpPInvokeGenerator phase, so that the external includes could match those remaps.

I'll double check when I'm back at my computer, but IIRC the size of the resulting winmd barely changes. The attributes are small and only added to the struct declarations.

sotteson1:
> > Why do we need this? We do a lot of renames. It seems like this is going to add a lot of bloat to the metadata. The whole "tag" notion is an old Win32 holdover. The "tag" versions of the struct names are nearly never used in real code.
> 
> The SDK went in a different direction than I expected by redistributing the rsp files in the NuGet. I initially made that change thinking that supporting #66 with only the winmd file would need to know about any remapping that happened in the ClangSharpPInvokeGenerator phase, so that the external includes could match those remaps.
> 
> I'll double check when I'm back at my computer, but IIRC the size of the resulting winmd barely changes. The attributes are small and only added to the struct declarations.
> 
> TL;DR; Turns out it's not required, but it might help simplify the SDK or enable other tools to reuse the winmd without the rsp files used to build it.

I would really like the change removed that has the winmd generator depend on the scraper .rsp file. I don't want the .winmd generator to have to know about the initial remaps. The initial .rsp files are absolutely necessary for the scraper to work on external projects so that things get mapped in a consistent way.

wravery:
> I'll double check when I'm back at my computer, but IIRC the size of the resulting winmd barely changes. The attributes are small and only added to the struct declarations.

Just to close the loop, the total size actually seems to have gone down from whenever the baseline was last generated:
```cmd
> ls .\scripts\BaselineWinmd\

    Directory: ...\win32metadata\scripts\BaselineWinmd

Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
-a---           6/13/2021  3:43 PM           6656 Windows.Win32.Interop.dll
-a---           6/26/2021  3:59 PM       13307904 Windows.Win32.winmd

> ls .\bin

    Directory: ...\win32metadata\bin

Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
d----           6/28/2021 11:26 AM                release
-a---           6/28/2021 11:25 AM           7680 Windows.Win32.Interop.dll
-a---           6/28/2021 11:50 AM       13294080 Windows.Win32.winmd
```
...and here's the WebView2 winmd:
```cmd
> ls .\sources\msbuild\WebView2SampleDebug\bin\

    Directory: ...\win32metadata\sources\msbuild\WebView2SampleDebug\bin

Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
d----           6/26/2021  4:31 PM                Debug
-a---           6/28/2021 11:54 AM          38400 Microsoft.Web.WebView2.winmd
```

> > TL;DR; Turns out it's not required, but it might help simplify the SDK or enable other tools to reuse the winmd without the rsp files used to build it.
> 
> I would really like the change removed that has the winmd generator depend on the scraper .rsp file. I don't want the .winmd generator to have to know about the initial remaps. The initial .rsp files are absolutely necessary for the scraper to work on external projects so that things get mapped in a consistent way.

OK, I can do that.

wravery:
> I would really like the change removed that has the winmd generator depend on the scraper .rsp file. I don't want the .winmd generator to have to know about the initial remaps. The initial .rsp files are absolutely necessary for the scraper to work on external projects so that things get mapped in a consistent way.

Latest force push removes that merge commit. For the most accurate comparison, here's the winmd size with just the remaining changes:
```cmd
> ls bin

    Directory: ...\win32metadata\bin

Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
d----           6/28/2021  1:11 PM                release
-a---           6/28/2021  1:11 PM           7680 Windows.Win32.Interop.dll
-a---           6/28/2021  1:35 PM       13293568 Windows.Win32.winmd
```
So overall adding those attributes would have added 512 bytes to the winmd. There must be some pretty impressive compression just to get the string literals down to that little, or maybe there's a lot of padding in the format already that let it fit in space that was already there.

sotteson1:
Thanks for doing this @wravery! I'm glad you've got it working for webview2.

sotteson1 (approved):

PR #570, opened 2021-07-07
MapViewOfFile should return a handle type

### LINKED ISSUE #561: MapViewOfFile should generate an overload returning SafeMemoryMappedViewHandle
## Actual behavior

`MapViewOfFile` only generates an overload returning `void*`. 

## Expected behavior

A separate overload, e.g. a generated `MapViewOfFile_SafeHandle` method, should return `SafeMemoryMappedViewHandle`.

https://github.com/dotnet/runtime/blob/57bfe474518ab5b7cfe6bf7424a79ce3af9d6657/src/libraries/Common/src/Interop/Windows/Kernel32/Interop.MapViewOfFile.cs#L12-L18

https://referencesource.microsoft.com/#System.Core/Microsoft/Win32/SafeHandles/SafeMemoryMappedViewHandle.cs,fbe85260617a73ea

## Repro steps

1. `NativeMethods.txt` content:
```
MapViewOfFile
```

2. `NativeMethods.json` content (if present):

N/A

3. Any of your own code that should be shared?

### Context

- CsWin32 version: `0.1.506-beta`
- Win32Metadata version (if explicitly set by project): N/A
- Target Framework: net45
- `LangVersion` (if explicitly set by project): `9`

mikebattista:
Fixes #561


AArnott:
@mikebattista tip: Use a descriptive title for PRs and make sure you mention the issue # in the PR description. In my list of PRs to review, "Fixes #561" is not a great title, particularly since the 561 in that isn't clickable, making it hard to find out what it actually fixes.

generation/emitter/remap.rsp
```diff
@@ -244,6 +244,16 @@ FindFirstVolumeMountPointW::return=FindVolumeMointPointHandle
...
+MapViewOfFile::return=MemoryMappedViewHandle
+MapViewOfFile3::return=MemoryMappedViewHandle
+MapViewOfFile3FromApp::return=MemoryMappedViewHandle
+MapViewOfFileEx::return=MemoryMappedViewHandle
+MapViewOfFileExNuma::return=MemoryMappedViewHandle
+MapViewOfFileFromApp::return=MemoryMappedViewHandle
+MapViewOfFileNuma2::return=MemoryMappedViewHandle
+UnmapViewOfFile::lpBaseAddress=MemoryMappedViewHandle
+UnmapViewOfFile2::BaseAddress=MemoryMappedViewHandle
+UnmapViewOfFileEx::BaseAddress=MemoryMappedViewHandle
```
AArnott, line 256: I wonder what the complications are from having two extra methods that release the handle. The RAIIRelease attribute will only point at the first one. So if the second and third would only be called manually, where would the caller have obtained the handle? I wonder if we should leave them accepting `IntPtr` instead of the new handle struct if that's the type that callers are likely to have.
I've never used any of these APIs, so it would be great to get the opinion of someone familiar with the area.

AArnott (approved):

generation/emitter/remap.rsp
mikebattista: You're supposed to use the UnmapViewOfFile that matches the MapViewOfFile API you called.

This one is interesting which is why I submitted a PR for review instead of just submitting this.

AArnott: Then I suppose we may need 3 handle types. And as the 2 and Ex methods take more than one parameter, they'll require special recognition by the projection so they know what to pass as the other argument(s).

mikebattista: Actually looks like the docs say you can mix and match a bit: https://docs.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-unmapviewoffile#parameters. If you need the extra flags depending on the scenario, you can unmap with -Ex.

What does .NET do here? Do they just call the base unmap function regardless of how the handle was acquired?

AArnott: .NET only exposes one SafeHandle AFAIK, and it uses the simply-named release function. But that's to say nothing of which method was used to obtain the handle in the first place, as .NET doesn't expose it.

riverar:
This PR is a bit stale, any updates?

PR #573, opened 2021-07-13
Make .zip, .winmd use git lfs

sotteson1:
Users will need to follow the instructions here to set up git lfs for their account: https://git-lfs.github.com/


AArnott:
Should your PR description be added as a note in the README file?

PR #597, opened 2021-08-01
Add null checks to ScrapeHeaders, ScanLibs to support no-lib scenarios

riverar:
This change adds null checks around the response files collection in the `ScrapeHeaders` task and libs property in the `ScanLibs` task. Without these checks, the tasks fail in scenarios where no applicable static libs were found and/or no libs were specified for scaping operations.


sotteson1 (approved):

sotteson1:
Thanks for the fix!

PR #622, opened 2021-08-29
Add a new HostCompute partition to contain the HCS APIs (CoreCompute.dll, CoreStorage.dll)

zhuman:
I've defined a new Windows.Win32.System.HostCompute partition for the [Host Compute System APIs](https://docs.microsoft.com/en-us/virtualization/api/hcs/overview), which were previously missing from win32metadata. I've included the necessary changes to generate RAII handle types in downstream projections. I performed basic testing of the projection by using it in a project.

In this change I also fixed a couple of minor issues found when running against the latest Windows 11 SDK. In particular, we need the newest ClangSharpPInvokeGenerator tool installed. I also found that ClangSharp would fail on char16_t usages in the Intl partition, so I've typedef'ed it to wchar_t.


generation/scraper/Partitions/Intl/main.cpp
```diff
@@ -1,6 +1,8 @@
 #define SECURITY_WIN32 // For sspi.h
 #define QCC_OS_GROUP_WINDOWS
 
+#define char16_t wchar_t
```
riverar, line 4: Nit: Can we `#include <uchar.h>` instead?

riverar (approved):

scripts/CommonUtils.ps1
```diff
@@ -64,7 +64,7 @@ function Install-DotNetTool
 
 function Install-BuildTools
 {
-    Install-DotNetTool -Name ClangSharpPInvokeGenerator -Version 11.0.0-beta3
+    Install-DotNetTool -Name ClangSharpPInvokeGenerator -Version 12.0.0-beta2
```
sotteson1, line 67: Please don't update this. I would need to make sure this doesn't regress anything before we update it.

sotteson1:
Hi @zhuman, I recently refactored how the partitions and .rsp files are laid out. I went ahead and took your changes for the HostCompute partition and integrated them into the master branch. Thanks a lot for helping us get these APIs.

zhuman:
Thanks a bunch!

PR #629, opened 2021-09-01
Create msbuild ProjectionMetadataWinmd item in package import

AArnott:
Instead of setting the `MicrosoftWindowsSdkWin32MetadataBasePath` msbuild property, which is a scalar and only allows one assignment, we now add to an msbuild item list so that projections (e.g. CsWin32) may consume many such metadata winmds in one build.

https://github.com/microsoft/CsWin32/pull/386 will consume this change.


sotteson1 (approved):

AArnott:
@sotteson1 The pipeline failed with a package push failure. Why are PR builds pushing packages anyway? And if it's intentional, can you look into the failure?

PR #630, opened 2021-09-01
Build authoring cleanup

AArnott:
- Lots of build authoring cleanup
  * Consolidate nuget.config files wherever possible.
  * Consolidate msbuild sdk versions into a global.json file.
  * Consolidate definitions of OutputPath in msbuild projects.
  * Remove/fix msbuild properties that were inconsistently or redundantly applied in project files.
  * Move Project Sdk attributes from Sdk.props to each project that imported it.
  * Remove a bunch of 1-project solution files in favor of a new BuildTools.sln that contains them all.
- Build the metadata package to the same path as docs by default.
- Suppress some pack warnings
- Delete obsolete sample pack script
- Add .editorconfig and apply to msbuild files
- Don't launch debugger during build unless an environment variable is set


sotteson1:
Thanks a ton for all the great fixes @AArnott! I learned a lot from the improvements you've made to the project.

PR #649, opened 2021-09-14
Fix metadata SDK project, change samples to build against built nuget packages

sotteson1:
Adding additional fixes to @AArnott's original changes.


AArnott (approved):

AArnott:
> error : The local source 'D:\a\1\s\bin\Packages\Release\NuGet' doesn't exist.

It looks like if you specify a local path in nuget.config, the directory had better exist. Maybe you should create it at the start of the pipeline.

sotteson1:
> > error : The local source 'D:\a\1\s\bin\Packages\Release\NuGet' doesn't exist.
> 
> It looks like if you specify a local path in nuget.config, the directory had better exist. Maybe you should create it at the start of the pipeline.

I'm working on a fix.

PR #654, opened 2021-09-15
Fixes #644 - Specify version of Microsoft.Build.NoTargets when doing Import

### LINKED ISSUE #644: winmdgenerator: 0.2.185-preview nuget is broken
MetadataTasks.dll is indeed missing https://nuget.info/packages/Microsoft.Windows.WinmdGenerator/0.2.185-preview

```
Sdk.targets(19,5): error MSB4062: The "MetadataTasks.CompileIdls" task could not be loaded from the assembly
C:\Users\Rafael\.nuget\packages\microsoft.windows.winmdgenerator\0.2.185-preview\tools\net5.0\MetadataTasks.dll.

Could not load file or assembly
'C:\Users\Rafael\.nuget\packages\microsoft.windows.winmdgenerator\0.2.185-preview\tools\net5.0\MetadataTasks.dll'.

The system cannot find the file specified. Confirm that the <UsingTask> declaration is correct, that the assembly
and all its dependencies are available, and that the task contains a public class that implements
Microsoft.Build.Framework.ITask.
```

sotteson1:
We need to specify a version of Microsoft.Build.NoTargets when we Import it in the props and targets. This didn't fail the build because we have a global.json with Microsoft.Build.NoTargets in it, which most users won't have.

### DIFF
--- a/sources/GeneratorSdk/sdk/sdk.props
+++ b/sources/GeneratorSdk/sdk/sdk.props
@@ -4,7 +4,7 @@
     <MicrosoftNoTargetsSdkImported Condition="'$(UsingMicrosoftNoTargetsSdk)' != 'true'">true</MicrosoftNoTargetsSdkImported>
   </PropertyGroup>
 
-  <Import Condition="'$(UsingMicrosoftNoTargetsSdk)' != 'true'" Sdk="Microsoft.Build.NoTargets" Project="sdk.props"/>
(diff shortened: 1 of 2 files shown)


AArnott (approved):
Have we confirmed that msbuild will download the sdk automatically with this?

sotteson1:
> Have we confirmed that msbuild will download the sdk automatically with this?

Yes, I tried it to make sure.

PR #686, opened 2021-09-28
Get the docs package to work for lower TFM projects

AArnott:


sotteson1:
There's an issue with the constants scraper that caused this to fail the tests. I'll take a look but I'm going to merge the PR.

PR #690, opened 2021-09-29
Capture undecorated symbol names from dumpbin output

### LINKED ISSUE #689: winmdgenerator: x86, arm64 libs are not scanned correctly resulting in missing metadata
MetadataTasks scans import lib dumpbin output for function names. It's currently pulling out `Symbol name`, which may be decorated, depending on target platform (e.g. _Function@0 on x86, Function on x64, etc.). This should instead read the undecorated `Name` field.

https://github.com/microsoft/win32metadata/blob/832ff5798733fc292c3bdbc73c8ae81e7a0cda4b/sources/GeneratorSdk/MetadataTasks/ScanLibs.cs#L35

riverar:
Fixes #689.

Alters the regular expression to instead capture undecorated symbol names.


riverar:
Closing, we don't need this anymore!

PR #692, opened 2021-09-30
Improving build to only scan non-arch-specific partitions once

### LINKED ISSUE #640: RtlInstallFunctionTableCallback has questionable multi-arch definitions
```
Duplicated RtlAddFunctionTable
Duplicated RtlDeleteFunctionTable
Duplicated RtlInstallFunctionTableCallback
Duplicated RtlAddGrowableFunctionTable    
Duplicated RtlLookupFunctionEntry
Duplicated RtlVirtualUnwind
Duplicated ImageNtHeader
Duplicated ImageRvaToSection
Duplicated ImageRvaToVa
Duplicated JsCreateContext
Duplicated JsStartDebugging
Duplicated GetDeviceID
```

sotteson1:
* You can exclude partitions from cross-arch processing two ways:
1) Add ````<ExcludeFromCrossarch\>true</ExcludeFromCrossarch\>```` to a partition in a .proj file.
2) Add a property at the project level with a semicolon-separated list: ````<ExcludeFromCrossarch\>partition1;partition2</ExcludeFromCrossarch\>````

The build system will detect if a partition doesn't need cross-arch processing and will let you know what to add to your proj file.

* Fixes #640


riverar:
Wouldn't most of the excluded partitions need cross arch to minimally handle structure packing differences?

sotteson1:
The majority of partitions scanned for all three architectures produce the exact same .cs file. That's why most structs/functions don't have the attribute indicating there are architectural differences. Most structs/functions are the same for all three. The build system will now detect if scanning x86 and arm64 for a given partition was essentially a no-op and warn you. Then you can add the partition to the list that don't need to be scanned for all three architectures.

PR #696, opened 2021-10-06
Lib scrape fixes

### LINKED ISSUE #689: winmdgenerator: x86, arm64 libs are not scanned correctly resulting in missing metadata

sotteson1:
* Fixes #689 - Fixes lib scraping issues
* Add tests around new lib scraping code
* Refactor how tests are called in DoAll.bat

These changes will make it easier to move to a new WinSDK, which I hope to do soon.


sources/MetadataUtils/LibScraper.cs
```diff
@@ -0,0 +1,266 @@
...
+                if (importHeader.ImportType != ImportType.Code)
+                {
+                    continue;
+                }
+
+                string symbolName = ReadNullTermString(libReader);
+                string dllName = ReadNullTermString(libReader);
+                string procName = symbolName;
+
+                // If it's x86, try to undecorate it
```
riverar, line 110: @sotteson1 What about x64/arm64ec though?

PR #698, opened 2021-10-14
Fixes #693 - Add MSDelta APIs

### LINKED ISSUE #693: Missing MSDelta methods
Are the [msdelta](https://docs.microsoft.com/en-us/windows/win32/devnotes/msdelta) methods expected to be included in the metadata? If not, where would we expect to find them?

sotteson1:


sotteson1:
Fixes #693

Issue #559, opened 2021-06-27
Add an attribute for identifying methods and functions that return alternative success codes


knopp:
For example `DoDragDrop` returns `DRAGDROP_S_DROP` on successful drop. However there doesn't seem to be any way to retrieve this value, since windows-rs immediately calls `ok()` on the `HRESULT`. And given that `DRAGDROP_S_DROP` doesn't have the error bit set it is not passed as error in the `Result`.

Is this an oversight?

knopp:
Didn't even know that was an option :) Sounds great, though I'm not clear on how this would work. I assume every call like this would need to be tagged manually, since there's no way to infer this from headers?

mikebattista:
You want an attribute that indicates that a function doesn't return S_OK on success?

damyanp:
This is for functions that return success HRESULTs that aren't S_OK.  So it might return S_OK, but they may also return S_FALSE, or some other >0 HRESULT to include additional data about the result.

Generally in projections we'd like to map HRESULTs to exceptions or whatever the idiomatic error handling scheme is for a language.  Most (if not all) of these assume that there only one possible way to indicate "success" and don't have a natural way to return "I succeeded, but....".  

We can build these things, but it's pretty unnatural and we think that for the vast majority of Win32/COM APIs it would be unnecessary, so we're looking for a way to only do this for the cases where we know that applications want to check for non-S_OK success codes.

mikebattista:
What would you like to see in the metadata specifically? I understand the scenario but I'm not clear how you want to model this.

damyanp:
I was thinking of a per-function attribute that indicates that applications might want to inspect successful HRESULTs.

mikebattista:
So just a boolean? And not a list of all possible success HRESULTs?

damyanp:
It might be possible for projections to make use of all the HRESULTs, but in practice I suspect most will opt to just pass the HRESULT back to the caller somehow, so a boolean would be fine.  @kennykerr might have thoughts on this.

kennykerr:
Just an attribute on functions/methods that return HRESULT. The attribute doesn't need any values (not even bool). It just signals to the languages that the HRESULT has alternate status codes.

mikebattista:
> Just an attribute on functions/methods that return HRESULT. The attribute doesn't need any values (not even bool). It just signals to the languages that the HRESULT has alternate status codes.

Ok. We can add support for an attribute but will likely need to crowd source where to apply it.

[...]

PR #758, opened 2021-11-25
Add additional AlternateSuccessCodes

dqtsth:
Hi i make search on google for "IEnum next site:https://docs.microsoft.com/en-us/windows/win32/" and found more methods with other return codes. Some return S_FALSE also IDirectorySearch and IADsPropertyList I find return other S_ codes like here https://docs.microsoft.com/en-us/windows/win32/adsi/generic-adsi-error-codes


mikebattista:
Thanks for providing the list.

Per https://github.com/microsoft/win32metadata/issues/559#issuecomment-908509892 we will be replacing the AlternateSuccessCodes attribute with just DllImport.PreserveSig. Once we do that, we'll update the design here to include the list.

kennykerr:
This is blocking customers: https://github.com/microsoft/windows-rs/issues/1371#issuecomment-988682189

> Per #559 (comment) we will be replacing the AlternateSuccessCodes attribute with just DllImport.PreserveSig. Once we do that, we'll update the design here to include the list.

Can we take this PR in the short term or are we planning on completing the replacement some time soon?

mikebattista:
I'll merge this in the short term.

mikebattista:
Merged. @sotteson1 will push a new metadata package to NuGet once the CI build finishes.

PR #789, opened 2022-01-31
Adjust PreserveSig emission

### LINKED ISSUE #559: Add an attribute for identifying methods and functions that return alternative success codes

riverar:
Fixes #559

Removes PreserveSig emission for all APIs by default. Adds PreserveSig emission for class methods explicitly marked with `[PreserveSig]` in `emitter.settings.rsp`. (Interface methods were already covered previously.)

Metadata diff is too large to be useful, as you can imagine.


sotteson1 (approved):

AArnott:
FWIW I feel PreserveSig should remain on standard C methods. Only the COM interface methods should lack preservesig because those are the ones we expect to be projected with a different signature.

kennykerr:
Hate to disagree but I've been applying transformations on methods and free functions alike and it would be a major step back not to. Developers love it.

AArnott:
@kennykerr I'm very curious now. Are you saying that win32 methods like CreateFile don't return values anymore? Methods that return hresult, win32 error codes, lstatus, etc... will throw instead of return error codes?

AArnott:
That makes sense. In CsWin32 we can deal with it either way, so if it's useful to you as-is, that's fine by me.

PR #819, opened 2022-02-23
Remap ICM mode constants to integers

### LINKED ISSUE #817: `SetICMMode` and `ICM_ON` use different types
These `ICM_*` constants are meant to be used with `SetICMMode` but they're the wrong type.

riverar:
Creates scoped enum for `ICM_` constants; realigns them to match `SetICMMode` input type expectations.

Fixes #817.

Metadata diff
```
Windows.Win32.UI.ColorSystem.ICM_MODE not found in 1st winmd
Windows.Win32.Graphics.Gdi.Apis.ICM_OFF not found in 2nd winmd
Windows.Win32.Graphics.Gdi.Apis.ICM_ON not found in 2nd winmd
Windows.Win32.Graphics.Gdi.Apis.ICM_QUERY not found in 2nd winmd
Windows.Win32.Graphics.Gdi.Apis.ICM_DONE_OUTSIDEDC not found in 2nd winmd
Windows.Win32.UI.ColorSystem.Apis.SetICMMode : mode...Int32 => ICM_MODE
```


mikebattista:
Should we create enums for SetICMMode and UpdateICMRegKeyA/W rather than loose constants? Are those constants used together anywhere or just in these corresponding functions?

riverar:
@mikebattista Ah yep, want to demote and scope these enums to `Windows.Win32.Graphics.Gdi.[...]` instead? Cursory glance indicates they're not used elsewhere.

mikebattista:
Actually it looks like UpdateICMRegKey already uses the ICM_COMMAND enum with just the values it needs.

We should create a separate enum for SetICMMode in the same namespace as these APIs which is Windows.Win32.UI.ColorSystem. I'd call it ICM_MODE. You can set the type to int in enums.json as well for this one. The namespace you shouldn't need to specify since it will inherit the namespace of SetICMMode by default.

Depending on where the SetICMMode constants are defined, you could use the autopopulate feature in enums.json, or if they're in a header with other ICM_ constants, you could manually define the members (just the names), and the values should be filled in by the constant scraper.

riverar:
@mikebattista Got it. Will do now.

mikebattista:
Thanks. Looks like the last build failed since the baseline wasn't updated. Be sure to update the baseline once you produce the new enum.

riverar:
@mikebattista All green. Is this what you had in mind?

mikebattista (approved):

mikebattista:
Looks perfect. Thanks!

Issue #25, opened 2020-11-14
Release methods need semantic understanding


AArnott:
In order to invoke the release methods properly, we need to understand whether they were successful. But they're all over the map on this. For example:

- `bool` [`FindClose`](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-findclose) returns non-zero (true) on success.
- `NTSTATUS` [`LsaDeregisterLogonProcess`](https://docs.microsoft.com/en-us/windows/win32/api/ntsecapi/nf-ntsecapi-lsaderegisterlogonprocess) returns zero (`STATUS_SUCCESS`) on success.
- `DWORD` [`WNetCloseEnum`](https://docs.microsoft.com/en-us/windows/win32/api/winnetwk/nf-winnetwk-wnetcloseenum) returns zero on success.
- `LSTATUS` [`RegCloseKey`](https://docs.microsoft.com/en-us/windows/win32/api/winreg/nf-winreg-regclosekey) returns zero on success.
- `byte` [`ClosePrivateNamespace`](https://docs.microsoft.com/en-us/windows/win32/api/namespaceapi/nf-namespaceapi-closeprivatenamespace) returns non-zero on success.

These is just a sampling. For each method, we need to know what constitutes success.

Robo210:
I believe the release methods will also need to understand what constitutes a "invalid" or sentinel value as well. Some APIs use NULL while others use INVALID_HANDLE_VALUE, and the two aren't always equivalent (but sometimes they are). If a safe handle needs to condition the close on whether the handle value is valid, it will need to know what an invalid handle looks like.

[WIL's unique_any](https://github.com/microsoft/wil/blob/86353933d5fba0caf61503c8b09e10e182dccdfb/include/wil/resource.h#L417) is a good example of all the data that's actually required to build a safe handle around the numerous Win32 handle types.

AArnott:
`ReleaseDC` is another odd case that takes two parameters.

PR #853, opened 2022-03-25
Add InvalidHandleValueAttribute support

riverar:
Adds a new `InvalidHandleValue` attribute (_AllowMultiple=true_) to attach to `NativeTypeDef` supporting structures (handles). This will be used by binding authors downstream to correctly implement invalid/valid handle semantics and to generate the appropriate checks before calling an associated `RAIIFree` function.

Also includes a minor refactor of the `NativeTypedefStructsCreator` class to make it testable. (Tests added.)

Examples:
```csharp
[RAIIFree("FindVolumeClose")]
[InvalidHandleValue(-1)]
[NativeTypedef]    
public struct FindVolumeHandle
{
   public IntPtr Value;
}

[RAIIFree("UnhookWinEvent")]
[InvalidHandleValue(0)]
[InvalidHandleValue(-1)]
[NativeTypedef]
public struct HWINEVENTHOOK
{
   public IntPtr Value;
}
```

Closes: #25, microsoft/windows-rs#1630


AArnott:
> AllowMultiple=false

Shouldn't we allow multiple values? Quite a few Windows handles consider 0 or -1 to both be invalid. So much so that .NET even defines [a SafeHandle type](https://apisof.net/catalog/d2b0fb89-ca33-0d0e-020f-d1eaa9adf4fd) to represent it.

I think Windows documents a bunch of handles as invalid with -1 as a value. Why then did you choose to express this as a `ulong`, which forces a representation that doesn't resemble the docs and creates an very unrecognizable value? Why not `long`?

riverar:
Any additional feedback? @AArnott @sotteson1 @mikebattista

sotteson1 (approved):
Looks great! Thanks @riverar.

PR #859, opened 2022-03-29
Add InvalidHandleValues to all handle types

riverar:
First attempt at defining invalid handle values for all the handle types in metadata. We'll probably be revisiting this a few more times before we reach a sweet spot.

I considered attaching default `InvalidHandleValue` attributes to all generated typedef structs, but that created an awkward authoring experience for `autoTypes.json` and I found myself _undoing_ the defaults in a non-trivial amount of cases. So I decided to go the explicit route instead. Thoughts welcome, of course.

Generally speaking, ...
* non-handles (e.g. `BSTR`) don't have `InvalidHandleValue` attributes
* the base `HANDLE` type declares `-1` and `0` as invalid
* specialized handle types mostly just behave as `HANDLE`s
* a few specialized handle special cases (e.g. HWND, HDDEDATA, UIntPtr-based handles) were handled


kennykerr:
Looks great - can't wait to try this!

sotteson1:
I would have thought you could leave off the empty InvalidHandleValue arrays in the json. Is that what you meant by undoing the defaults? When I see empty json arrays it makes me think the code backing it could be improved to do the right thing if the value isn't specified.

riverar:
@sotteson1 By "undoing the defaults" I just meant I initially baked in `0`/`-1` as defaults but then had to do a bunch of work to _undo_ that for things like BSTR, LPARAM, HWND, etc. So I abandoned that approach.

But you're right, I can remove the empty arrays, will commit that (and an updated winmd) in a second.

sotteson1 (approved):

Issue #1021, opened 2022-07-30
All `DllImport` attributes should have fully qualified library name


riverar:
Similar: #777, #714

mikebattista:
See comments in the reverted PR. The baseline winmd was corrupt so reverted the changes.

PR #1022, opened 2022-07-30
Add .dll suffix to remaining library mappings

riverar:
Removes remaining library name ambiguity in metadata. This makes it easier, for example, for downstream projections to create import libraries. (Libraries that had non-`.dll` suffixes, such as `.ax` and `.drv`, were already fully qualified in earlier commits.)

Fixes: #1021


riverar:
@AArnott This won't mess anything up on your end right? Just double-checking.

mikebattista:
> @AArnott This won't mess anything up on your end right? Just double-checking.

@AArnott could you please review?

mikebattista (changes requested):
Pending approval from @AArnott.

AArnott (approved):
Thanks for checking. I don't anticipate any problems. And if there are, I'm confident CsWin32 can be changed to accommodate them.

mikebattista (approved):

mikebattista:
The baseline winmd from this was getting flagged by Windows Defender with a Trojan, so I reverted main.

image

You should take a look at what's going on. You'll need to rebuild and resubmit this change.

PR #1047, opened 2022-08-15
Changes R2_MODE and TEXT_ALIGN_OPTIONS

elachlan:
Fixes: 
- https://github.com/microsoft/win32metadata/issues/1036
- https://github.com/microsoft/win32metadata/issues/1045
```
Windows.Win32.Graphics.Gdi.TEXT_ALIGN_OPTIONS :  => [Flags]
Windows.Win32.Graphics.Gdi.Apis.GetROP2 : return...Int32 => R2_MODE
Windows.Win32.Graphics.Gdi.Apis.GetTextAlign : return...UInt32 => TEXT_ALIGN_OPTIONS
```

### DIFF
--- a/generation/WinSDK/enums.json
+++ b/generation/WinSDK/enums.json
@@ -3861,6 +3861,7 @@
     "type": "uint",
     "namespace": "Windows.Win32.Graphics.Gdi",
     "name": "TEXT_ALIGN_OPTIONS",
+    "flags": true,
     "members": [
       {
         "name": "TA_NOUPDATECP",
@@ -3936,6 +3937,10 @@
       {
         "struct": "RICHEDIT_IMAGE_PARAMETERS",
         "field": "Type"
+      },
+      {
+        "method": "GetTextAlign",
+        "parameter": "return"
       },
       {
         "method": "SetTextAlign",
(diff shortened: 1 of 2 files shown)


elachlan:
CC: @lonitra

generation/WinSDK/enums.json
```diff
@@ -3861,6 +3861,7 @@
     "type": "uint",
     "namespace": "Windows.Win32.Graphics.Gdi",
     "name": "TEXT_ALIGN_OPTIONS",
```
elachlan, line 3863: Should `TEXT_ALIGN_OPTIONS` use `autoPopulate`? Would that mean we would only need to define `TA_MASK` in `members`?

sotteson1: Yes, you could do that, but I would only change it if something is broken. If it's already working, I would leave it as is.

elachlan: I will leave it as is then.

elachlan:
Cc: @AArnott

sotteson1:
The new system gets rid of the baseline .winmd, and instead updates a text file that will include the diff of your changes. (The output of the build is meant to help you automate your additions to this file.) Sorry to randomize your PR but I think this system will be a lot nicer for merges in the future.

riverar (approved):

sotteson1 (approved):
Thanks for all the contributions you've been giving us!

PR #1058, opened 2022-08-18
Fix build error

elachlan:
Fix the build error and have a nice output.


elachlan:
@sotteson1 this feels a bit nicer than `write-error`.

sotteson1:
I purposely changed this to an error, because I want to be clear the build is failing until the script is run to update the known exceptions.

elachlan:
Maybe we can highlight it red then instead? This error makes it seem like the script is broken.

riverar:
`Write-Error` writes to the error stream amongst other things. Feel free to propose an alternative that preserves this behavior or lets close the PR and move on.

PR #1061, opened 2022-08-18
Adding a script to help set up the environment 

chenss3:
Install-AllSoftware.ps1 uses WinGet to automate the software installation listed in the Contributing guide. Hopefully, this will make environment set up and contributing easier.


scripts/Install-AllSoftware.ps1
```diff
@@ -0,0 +1,24 @@
+# 1. Make sure the Microsoft App Installer is installed:
+#    https://www.microsoft.com/en-us/p/app-installer/9nblggh4nns1
+# 2. Run this script as administrator. If running scripts is blocked, you can temporarily unblock them by running 
+	Set-ExecutionPolicy -ExecutionPolicy Unrestricted -Scope Process
+
+Write-Output "Installing apps"
+$apps = @(
+	@{name = "GitHub.GitLFS" },
```
elachlan, line 8: Is GitLFS still required?

CONTRIBUTING.md
```diff
@@ -37,6 +37,9 @@ contact [opencode@microsoft.com](mailto:opencode@microsoft.com) with any additio
 * Install [PowerShell 7](https://aka.ms/powershell-release?tag=stable)
 * Install [.NET (SDK)](https://dotnet.microsoft.com/download/dotnet)
 * Install [ILSpy](https://github.com/icsharpcode/ILSpy/releases/latest)
+
+Run `./scripts/Install-AllSoftware.ps1 ` to install all of the aforementioned software. To use this script, make sure that the [Microsoft App Installer](https://www.microsoft.com/en-us/p/app-installer/9nblggh4nns1) is installed, the script is run as an administrator, and running scripts is not blocked.
```
riverar, line 41: Remove hardcoded English locale.

```suggestion
Run `./scripts/Install-AllSoftware.ps1 ` to install all of the aforementioned software. To use this script, make sure that the [Microsoft App Installer](https://www.microsoft.com/p/app-installer/9nblggh4nns1) is installed, the script is run as an administrator, and running scripts is not blocked. 
```

scripts/Install-AllSoftware.ps1
```diff
@@ -0,0 +1,24 @@
+# 1. Make sure the Microsoft App Installer is installed:
+#    https://www.microsoft.com/en-us/p/app-installer/9nblggh4nns1
```
riverar, line 2: Remove hardcoded English locale.

```suggestion
#    https://www.microsoft.com/p/app-installer/9nblggh4nns1
```

scripts/Install-AllSoftware.ps1
```diff
@@ -0,0 +1,24 @@
...
+# 2. Run this script as administrator. If running scripts is blocked, you can temporarily unblock them by running 
+	Set-ExecutionPolicy -ExecutionPolicy Unrestricted -Scope Process
```
riverar, line 4: Microsoft needs to sign this script then we can rely on default RemoteSigned behavior and delete step 2 here.

scripts/Install-AllSoftware.ps1
```diff
@@ -0,0 +1,24 @@
...
+	@{name = "GitHub.GitLFS" },
+	@{name = "icsharpcode.ILSpy" },
+	@{name = "Microsoft.DotNet.SDK.6" },
+   	@{name = "Microsoft.VisualStudioCode" }
+);
+Foreach ($app in $apps) {
+    $listApp = winget list --exact -q $app.names
+    if (![String]::Join("", $listApp).Contains($app.name)) {
+        Write-host "Installing: " $app.name
+        winget install -e -h --accept-source-agreements --accept-package-agreements --id $app.name
```
riverar, line 17: I don't think Microsoft can/should automatically accept binding agreements on behalf of the user. We'll have to remove those flags or somehow show all agreements upfront (which certainly diminishes the value of this script).

scripts/Install-AllSoftware.ps1
```diff
@@ -0,0 +1,24 @@
...
+   	@{name = "Microsoft.VisualStudioCode" }
+);
+Foreach ($app in $apps) {
+    $listApp = winget list --exact -q $app.names
+    if (![String]::Join("", $listApp).Contains($app.name)) {
+        Write-host "Installing: " $app.name
+        winget install -e -h --accept-source-agreements --accept-package-agreements --id $app.name 
+    }
+    else {
+        Write-host "Skipping: " $app.name " (already installed)"
```
riverar, line 20: As a good practices measure, we should lean on `Write-Verbose` or `Write-Output` here instead. https://devblogs.microsoft.com/scripting/understanding-streams-redirection-and-write-host-in-powershell/

scripts/Install-AllSoftware.ps1
denelon: They aren't necessary at all since none of these packages are coming from the Microsoft Store "msstore" source in WinGet. Since you have the exact PackageIdentifier values called out, the only argument that could be useful here is the -h to get a truly silent install experience as opposed to "silent with progress" which is the default.

denelon: It might be worth adding "-s winget" to make sure you are using the community repository.

You could also check whether the "winget" source is available via `winget source list` in the event group policy may have restricted or removed the source.

chenss3: That's a good point, Rafael, and thank you for the insights, Demitrius. Because all of my apps are coming fromthe community repository and not MSStore, I got rid of all the agreement acceptance tags. I also got rid of the other "useless" tags. Changes should be reflected in my latest push.

scripts/Install-AllSoftware.ps1
chenss3: Makes sense! Omitted the verbatim in my comments and will look more into this.

riverar: (If the other developer scripts are unsigned, then disregard.)

scripts/Install-AllSoftware.ps1
sotteson1: We shouldn't need it anymore since we no longer have any GitLFS files.

scripts/Install-AllSoftware.ps1
chenss3: The other ps scripts are unsigned in this proj so I'll keep this one as unsigned too, but it's definitely a good thing to look into later. Will probably create an issue for it soon.

scripts/Install-AllSoftware.ps1
```diff
@@ -0,0 +1,23 @@
...
+Write-Output "Installing apps"
+$apps = @(
+	@{packageID = "icsharpcode.ILSpy" },
+	@{packageID = "Microsoft.DotNet.SDK.6" },
+   	@{packageID = "Microsoft.VisualStudioCode" }
+);
+Foreach ($app in $apps) {
+    $listApp = winget list --exact -q $app.packageIDs
+    if (![String]::Join("", $listApp).Contains($app.packageID)) {
+        Write-output "Installing: " $app.packageID
```
riverar, line 15: Capitalization nit.
```suggestion
        Write-Output "Installing: " $app.packageID
```

scripts/Install-AllSoftware.ps1
```diff
@@ -0,0 +1,23 @@
...
+        winget install -h $app.packageID -s winget 
+    }
+    else {
+        Write-output "Skipping: " $app.packageID " (already installed)"
```
riverar, line 19: Instead of outputting 3 separate strings into the pipeline, we can use the [subexpression operator](https://docs.microsoft.com/powershell/module/microsoft.powershell.core/about/about_operators?view=powershell-7.2#subexpression-operator--) to get that package ID into the string.

```suggestion
        Write-Output "Skipping: $($app.packageID) (already installed)"
```

riverar (approved):
Just minor non-blocking nits.

scripts/Install-AllSoftware.ps1
chenss3: Makes sense! Made this modification to the other output line too.

riverar (dismissed):

scripts/Install-AllSoftware.ps1
```diff
@@ -0,0 +1,23 @@
+# 1. Make sure the Microsoft App Installer is installed:
+#    https://www.microsoft.com/p/app-installer/9nblggh4nns1
+# 2. Run this script as administrator. If running scripts is blocked, you can temporarily unblock them by running 
+#    Set-ExecutionPolicy -ExecutionPolicy Unrestricted -Scope Process
```
BenJKuhn, line 4: if you can, it would be good to check these two conditions in code and give an appropriate error message.

chenss3: Good idea! My latest push now checks and outputs appropriate error msgs.

scripts/Install-AllSoftware.ps1
```diff
@@ -0,0 +1,47 @@
...
+    Foreach ($app in $apps) {
+    $listApp = winget list --exact -q $app.packageIDs
+    if (![String]::Join("", $listApp).Contains($app.packageID)) {
+        Write-Output "Installing: $($app.packageID)"
+        winget install -h $app.packageID -s winget 
+    }
+    else {
+        Write-Output "Skipping: $($app.packageID) (already installed)"
+    }
+    }
```
sotteson1, line 43: Nit: should be indented at same level as the matching else

scripts/Install-AllSoftware.ps1
```diff
@@ -0,0 +1,47 @@
...
+ }
+
+$policy = Get-ExecutionPolicy
+if ($policy -ne "Unrestricted"){
+    throw "Please set the execution policy to Unrestricted. You can temporarily unblock them by running
+    Set-ExecutionPolicy -ExecutionPolicy Unrestricted -Scope Process"
+}
+elseif (-not(Check-IsElevated)){
+    throw "Please run this script as an administrator"
+ }
```
sotteson1, line 25: Nit: move brace to match elseif

scripts/Install-AllSoftware.ps1
```diff
@@ -0,0 +1,47 @@
...
+    if (![String]::Join("", $listApp).Contains($app.packageID)) {
+        Write-Output "Installing: $($app.packageID)"
+        winget install -h $app.packageID -s winget 
+    }
+    else {
+        Write-Output "Skipping: $($app.packageID) (already installed)"
+    }
+    }
+
+    & "$PSScriptRoot\Get-VSPath.ps1" | Out-Null
```
sotteson1, line 45: What does this line do?

sotteson1 (changes requested):

scripts/Install-AllSoftware.ps1
chenss3: It will check to see if the user as Visual Studio, along with its required workloads, installed, and if not, it will prompt the user and makes use of the Install-VS.ps1 script we already have.

sotteson1 (approved):

PR #1062, opened 2022-08-19
Adding LOGFONTW Enums

### LINKED ISSUE #1060: Consider Updating LOGFONTW
[`LOGFONTW` docs](https://docs.microsoft.com/en-us/windows/win32/api/wingdi/ns-wingdi-logfontw) outline enum-like typed values for certain members of its struct e.g. `lfOutPrecision`,`lfQuality`. It would be nice to update the definition to return an enum type for these members for readability.
For example, with `lfOutPrecision`, something like:
```c#
public enum OUT_PRECIS : byte
{
    DEFAULT = 0,
    STRING = 1,
    CHARACTER = 2,
    STROKE = 3,
    TT = 4,
    DEVICE = 5,
    RASTER = 6,
    TT_ONLY = 7,
    OUTLINE = 8,
    SCREEN_OUTLINE = 9,
    PS_ONLY = 10
}
```

The docs also outline an `LF_FACESIZE`, which indicates the size of `lfFaceName`. Currently, the type of `lfFaceName` has a property `Length` which is essentially the `LF_FACESIZE`, but it would be nice to have this defined as an `internal const` in `LOGFONTW` to be able to reference this without having to first create one i.e. reference this value by calling `LOGFONTW.LF_FACESIZE`.

elachlan:
Fixes #1060

### DIFF
--- a/scripts/ChangesSinceLastRelease.txt
+++ b/scripts/ChangesSinceLastRelease.txt
@@ -174,3 +174,52 @@ Windows.Win32.UI.WindowsAndMessaging.Apis.UpdateLayeredWindow : crKey...UInt32 =
 Windows.Win32.UI.WindowsAndMessaging.UPDATELAYEREDWINDOWINFO.crKey...System.UInt32 => Windows.Win32.Foundation.COLORREF
 Windows.Win32.UI.Controls.DTTOPTS_FLAGS added
 Windows.Win32.UI.Controls.DTTOPTS.dwFlags...System.UInt32 => Windows.Win32.UI.Controls.DTTOPTS_FLAGS
+# LOGFONT enum changes
+Windows.Win32.Graphics.Gdi.Apis.ANSI_CHARSET removed
+Windows.Win32.Graphics.Gdi.Apis.ARABIC_CHARSET removed
+Windows.Win32.Graphics.Gdi.Apis.BALTIC_CHARSET removed
+Windows.Win32.Graphics.Gdi.Apis.CHINESEBIG5_CHARSET removed
+Windows.Win32.Graphics.Gdi.Apis.DEFAULT_CHARSET removed
+Windows.Win32.Graphics.Gdi.Apis.EASTEUROPE_CHARSET removed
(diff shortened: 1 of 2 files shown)


elachlan:
CC: @lonitra

scripts/ChangesSinceLastRelease.txt
```diff
@@ -172,3 +172,52 @@ Windows.Win32.UI.WindowsAndMessaging.Apis.SetLayeredWindowAttributes : crKey...U
...
+Windows.Win32.Graphics.Gdi.Apis.FW_EXTRALIGHT removed
+Windows.Win32.Graphics.Gdi.Apis.FW_HEAVY removed
+Windows.Win32.Graphics.Gdi.Apis.FW_LIGHT removed
+Windows.Win32.Graphics.Gdi.Apis.FW_MEDIUM removed
+Windows.Win32.Graphics.Gdi.Apis.FW_NORMAL removed
+Windows.Win32.Graphics.Gdi.Apis.FW_REGULAR removed
+Windows.Win32.Graphics.Gdi.Apis.FW_SEMIBOLD removed
+Windows.Win32.Graphics.Gdi.Apis.FW_THIN removed
+Windows.Win32.Graphics.Gdi.Apis.FW_ULTRABOLD removed
+Windows.Win32.Graphics.Gdi.Apis.FW_ULTRALIGHT removed
```
riverar, line 196: This diff output doesn't indicate if all these members survived and were added to Windows.Win32.Graphics.Gdi.FONT_WEIGHT. So fingers crossed! (Not a blocker)

riverar (dismissed):

scripts/ChangesSinceLastRelease.txt
elachlan: I will load it up in ilspy and confirm. I don't want to break anything.

elachlan: Verified.

sotteson1 (approved):
Thanks for the contribution!

PR #1204, opened 2022-08-31
Fix duplicate guid constants and structs

### LINKED ISSUE #1202: Duplicate constants
I thought I'd give #1201 a try and while I see the calling conventions in metadata 🥳 I also see duplicate constants. For example, `CODECAPI_ALLSETTINGS` appears both as a constant inside the `Apis` class and as a `struct` of its own.

I just grabbed the internal Azure build but perhaps that's missing some post processing and I should just wait for the nuget release. Anyway, just wanted to make sure you're aware of this as it prevents me from ingesting this build.

sotteson1:
Get rid of the structs that duplicate some constants.
Convert guid-only structs to guid constants

Fixes #1202


kennykerr:
> Improve the tests to count an empty struct with a guid attribute to be a constant.

Better still would be to actually turn "an empty struct with a guid attribute" into a constant so that we have a consistent representation for such constants. That way I don't have to harmonize this on my end. 😜

sotteson1:
> > Improve the tests to count an empty struct with a guid attribute to be a constant.
> 
> Better still would be to actually turn "an empty struct with a guid attribute" into a constant so that we have a consistent representation for such constants. That way I don't have to harmonize this on my end. 😜

I'm fine with that too. I thought having a guid on a struct meant it represented something that could be CoCreated, whereas a guid constant could mean lots of other things. But I'm happy to pile them all together as constants.

kennykerr:
Yep, there's no meaningful distinction so that would be great.

PR #1228, opened 2022-09-08
Add schema for enums.json

### LINKED ISSUE #1226: Add json schema for `enums.json`
This will help with autocomplete and validation.
Example Schema from CsWin32: https://github.com/microsoft/CsWin32/blob/main/src/Microsoft.Windows.CsWin32/settings.schema.json
Need to create a URL: `https://aka.ms/Win32MetaDataEnums.schema.json`
Point it to `Win32metadata\generation\WinSDK\enums.schema.json`

elachlan:
In doing this, we found several errors in the enums.json file, and corrections are included in this PR.

VS Code and Visual Studio will catch such errors and squiggle them and put them into the error list / problems pane to draw attention, like this:
image
or this:
image

Fixes #1226


elachlan:
We might need to do this in stages. With the schema file added first, then the enums.json updated with the reference to the schema URL created by one of the team members. `https://aka.ms/Win32MetaDataEnums.schema.json` for example pointing to `https://github.com/microsoft/win32metadata/generation/WinSDK/enums.schema.json`.

AArnott:
> We might need to do this in stages

Not at all. These json files can point to schemas at relative file locations. For example, [check this out](https://github.com/microsoft/CsWin32/blob/fef9088084dc1290fcced114f72d2b6dfbb8350d/test/GenerationSandbox.Tests/NativeMethods.json#L2).

elachlan:
Cool! I am learning more everyday! I'll sort it out now.

elachlan:
@AArnott, do you know of a way to specify the schema on the json doc for an array? I am guessing we need to turn it into an object with an items property to be able to specify the schema?

elachlan:
I think I worked it out. It seems to be applying the schema validation now. Build fails because of the deserialization. I think I have to change it here:
https://github.com/microsoft/win32metadata/blob/c24edb22bdc163cb7e2cf04d585de40eebff191c/sources/MetadataUtils/EnumObject.cs#L19-L22

elachlan:
Tried changing the deserialization and failed. Any tips would be fantastic.

AArnott:
Ah, I see the problem you're struggling with. I know of no way to place the `$schema` element when the top-level element is an array. I'll take a look.

AArnott:
I've got it working. I made a bunch of other changes to logging mechanisms to make this easier to diagnose, which I'll include in a separate PR. I'll try to push to the source branch of this PR with the fixes to the build.

AArnott:
I've pushed my fixes. In addition to fixing the build break, I enhanced the schema.json file to be more precise (forbidding extra properties, requiring _either_ a struct field _or_ a method parameter), which led to discovery of several bugs in the json file. 

I also reverted a `flags: true` change you made since that is irrelevant to this PR. If you meant to make that change, can you submit it separately?

I also reverted the indentation change to the enums.json file so that this PR can show the meaningful diff introduced to the file.

Thank you for putting this together, @elachlan!

AArnott (approved):

AArnott (approved):

AArnott:
@sotteson1 @mikebattista I recommend *squashing* this PR because the enums.json file has been totally rewritten, then mostly reverted. If merged, this PR will 'blame' me for nearly every line's existence. If squashed, only the lines with a net change will be 'blamed' on the PR author.
Alternatively, @elachlan can rebase/squash the commits and force push to this PR. If you do that, be sure to pull first, since I pushed commits to your PR.

mikebattista (approved):

mikebattista:
This all looks great. Thanks for implementing this!

elachlan:
@AArnott, thanks for taking this on! This is fantastic, it should make reviews a lot easier.

PR #1229, opened 2022-09-08
Logging improvements

AArnott:
I needed some of these changes to diagnose the enum parsing failures from #1228. Since they are overall goodness, I wanted to share.

-  Capture msbuild binlogs in local builds and in the pipeline
- Emit constants scraper errors in the console log
  Errors should be written to STDERR, and msbuild should invoke tools such that STDERR gets treated with high priority.
- Use powershell `Write-Host` and `Write-Error` as appropriate
    This avoids needing to use terminal escape codes to colorize output.
    This directs errors to the STDERR stream as appropriate.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -33,7 +33,7 @@ jobs:
       targetType: inline
       script: |
         .\scripts\Install-DotNetTool.ps1 -Name nbgv
-        nbgv cloud      
+        nbgv cloud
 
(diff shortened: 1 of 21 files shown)


mikebattista (approved):

mikebattista:
Thanks!

sotteson1:
Thanks @AArnott! All great changes.

PR #1250, opened 2022-09-14
Fix `DWriteCreateFactory` signature

### LINKED ISSUE #1248: `DWriteCreateFactory` has incorrect parameter type
Could I please request an example of how to use DirectWrite? I'm working with a Direct2D project, and although the direct2d sample has been very useful to me it doesn't draw any text. 

Specifically, I have this:
```
fn create_write_factory() -> Result<IUnknown> {
    unsafe { 
        let result = DWriteCreateFactory(DWRITE_FACTORY_TYPE_SHARED, &IDWriteFactory1::IID)?;
       
        Ok(result)
    }
}
```
And would like to return a `Result<IDWriteFactory1>`, but I'm not sure how to cast the `IUnknown`. The `D2D1CreateFactory` called in [the direct2d sample](https://github.com/microsoft/windows-rs/blob/master/crates/samples/direct2d/src/main.rs#L405) makes such a cast unnecessary, as the method is generic and returns ID2D1Factory1 without the need for a cast.

kennykerr:
Fixes #1248

### DIFF
--- a/generation/WinSDK/emitter.settings.rsp
+++ b/generation/WinSDK/emitter.settings.rsp
@@ -495,6 +495,7 @@ LoadCursorW::return=[DoNotRelease]
 GetModuleHandleA::return=[DoNotRelease]
 GetModuleHandleW::return=[DoNotRelease]
 IWeakReference::Resolve::objectReference=[ComOutPtr]void**
+DWriteCreateFactory::factory=[ComOutPtr]void**
 PCEVENT_DESCRIPTOR=[Const]
 PCCERT_CONTEXT=[Const]
 EXCEPTION_RECORD::ExceptionCode=NTSTATUS


mikebattista (approved):

kennykerr:
This was my first attempt at a PR. I tried fixing the build according to the Azure build failure message but running the `UpdateChangesSinceLastRelease` script failed for mysterious reasons. 

I'll leave this for now, but I'd suggest making this a little more discoverable especially since those that don't have access to the Azure build logs presumably have to guess why the build failed.

mikebattista:
What were the failures updating the baseline? That seems to be the only problem with the build is the baseline wasn't updated.

kennykerr:
The build long included this:

```
Testing .winmd succeeded
*** Comparing .winmd to last release
Unknown differences found:
Windows.Win32.Graphics.DirectWrite.Apis.DWriteCreateFactory : factory...IUnknown* => Void**
D:\a\1\s\scripts\CompareBinToLastRelease.ps1 : If all the differences are expected, please update the expected differences list:
.\scripts\UpdateChangesSinceLastRelease.ps1 '<add a reason here>'
```

So I tried to run `CompareBinToLastRelease` and, after enabling unsigned scripts 💀, it failed as follows:

```
D:\git\win32metadata>powershell scripts\CompareBinToLastRelease.ps1 'Fix DWriteCreateFactory signature'
Could not execute because the application was not found or a compatible .NET SDK is not installed.
```

mikebattista:
Looks like you ran the wrong script. Can you try running `.\scripts\UpdateChangesSinceLastRelease.ps1 '<add a reason here>'` per the build output?

If that fails the same way, you might be missing a development prerequisite.

kennykerr:
Ah, that explains why the error didn't look right. Here's the actual error I got:

```
D:\git\win32metadata>powershell scripts\UpdateChangesSinceLastRelease 'Fix DWriteCreateFactory signature'
D:\git\win32metadata\scripts\CompareBinToLastRelease.ps1 : A parameter cannot be found that matches parameter name
'LeafBase'.
At D:\git\win32metadata\scripts\UpdateChangesSinceLastRelease.ps1:9 char:1
+ & "$PSScriptRoot\CompareBinToLastRelease.ps1" -SkipInstallTools -Upda ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidArgument: (:) [CompareBinToLastRelease.ps1], ParameterBindingException
    + FullyQualifiedErrorId : NamedParameterNotFound,CompareBinToLastRelease.ps1
```

mikebattista:
Are you running the script in Windows PowerShell? You probably need to use PowerShell 7 / pwsh instead of powershell.

kennykerr:
Sorry, I don't know anything about powershell.

mikebattista:
Run `winget install --id Microsoft.PowerShell`. Then launch PowerShell 7 from Start and run the scripts there. You can also replace `powershell` in your command with `pwsh` once it is installed. That should also work.

kennykerr:
Next error:

image

mikebattista:
I would try running a clean build within PowerShell 7 with `DoAll.ps1 -Clean`, and then running the baseline update when it completes.

kennykerr:
I was trying to see how far I'd get without learning a whole new build system. Not that far as it turns out. Unfortunately I'm swamped on the Rust side so won't get to this.

mikebattista:
You need to have proper development prerequisites installed and need to be able to build the metadata to contribute.

kennykerr:
Yep, sorry for the noise.

mikebattista:
FYI your change was correct. You just need to get a full build running on your machine to contribute in the future.

Issue #92, opened 2021-01-11
Avoid API set names in metadata


kennykerr:
[The doc](https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/platform-apis/ms775098(v=vs.85)) says this function is exported from `urlmon.dll` but the metadata says `ext-ms-win-core-iuri-l1-1-0`. Even if that works up-level, we should prefer the original DLL names as that provides compatibility down-level and "just works" everywhere.

DaLynX:
I am having the same issue with `MiniDumpWriteDump` which is in `DbgHelp.dll` and documented as such but gets linked with `api-ms-win-core-debug-minidump-l1-1-0.dll`.

sotteson1:
The scraper is trying to use the non-api set whenever it exists. It looks like both of these methods have been fixed by that change. Let me know if you see others you think shouldn't be using an api set.

PR #1257, opened 2022-09-17
Map synchronization functions to `api-ms-win-core-synch-l1-2-0.dll`

ChrisDenton:
While win32metadata [prefers not to use API sets](https://github.com/microsoft/win32metadata/issues/92). the canonical lib for `WaitOnAddress` et al uses `api-ms-win-core-synch-l1-2-0.dll`. See also [the MSDN documentation](https://learn.microsoft.com/en-us/windows/win32/api/synchapi/nf-synchapi-waitonaddress). And using `vertdll` causes issues in github's CI with `windows-latest` (see https://github.com/microsoft/windows-rs/issues/2027).

---

For the sake of completeness I used `dumpbin /headers` to convert SDK lib files into human readable forms and grepped the libraries:
```text
> rg 'Symbol name  : WaitOnAddress' --before-context 1
windowscoreheadless_apiset.lib
43538-  DLL name     : api-ms-win-core-synch-l1-2-0.dll
43539:  Symbol name  : WaitOnAddress

windowscoreheadless.lib
112946-  DLL name     : api-ms-win-core-synch-l1-2-0.dll
112947:  Symbol name  : WaitOnAddress

WindowsApp_downlevel.lib
30464-  DLL name     : api-ms-win-core-synch-l1-2-0.dll
30465:  Symbol name  : WaitOnAddress

WindowsApp.lib
30464-  DLL name     : api-ms-win-core-synch-l1-2-0.dll
30465:  Symbol name  : WaitOnAddress

vertdll.lib
112-  DLL name     : vertdll.dll
113:  Symbol name  : WaitOnAddress

synchronization.lib
163-  DLL name     : api-ms-win-core-synch-l1-2-0.dll
164:  Symbol name  : WaitOnAddress

OneCore_apiset.lib
41129-  DLL name     : api-ms-win-core-synch-l1-2-0.dll
41130:  Symbol name  : WaitOnAddress

OneCoreUAP_apiset.lib
43919-  DLL name     : api-ms-win-core-synch-l1-2-0.dll
43920:  Symbol name  : WaitOnAddress

OneCore.lib
103619-  DLL name     : api-ms-win-core-synch-l1-2-0.dll
103620:  Symbol name  : WaitOnAddress

OneCoreUAP.lib
142648-  DLL name     : api-ms-win-core-synch-l1-2-0.dll
142649:  Symbol name  : WaitOnAddress

mincore.lib
51669-  DLL name     : api-ms-win-core-synch-l1-2-0.dll
51670:  Symbol name  : WaitOnAddress

mmos.lib
52943-  DLL name     : api-ms-win-core-synch-l1-2-0.dll
52944:  Symbol name  : WaitOnAddress
```

I think vertdll is a weird library to be using for this because it's mostly for enclave stuff and just so happens to export some other things (such as `memset` and `WaitOnAddress`).


kennykerr (approved):

riverar (changes requested):
Not sure I understand the issue fully. `vertdll.dll` exports this API on both Windows Server 2022 21H2 and Windows 11.

riverar:
OK `vertdll.dll` is indeed the wrong library for this. And the related issue is a result of Server nanoserver missing enclave APIs.  If we're going to remain consistent in our avoidance of API sets, shouldn't we set this to `kernelbase.dll` instead?

riverar (approved):

riverar:
This PR is missing the metadata change diff. Not sure why build is not failing as a result.

ChrisDenton:
My only issue with that is `kernelbase.dll` is essentially a private implementation detail in the sense that it doesn't appear in any of the public sdk import libraries. And we already do use a bunch of api sets so it wouldn't be inconsistent with what's already there. E.g. the imports right above these:

https://github.com/microsoft/win32metadata/blob/aed52b8e3b36ef452c62ec08f00c3cb18e7c3d15/generation/WinSDK/libMappings.rsp#L10805-L10878

kennykerr:
I don't mind using this for now, but we should figure out the story for these references. The loader ends up having to "fix" these all up to point back to the original DLLs, which seems wasteful and they're just cryptic and confusing. And it turns out the redirection feature the API sets were originally intended to provide was baked directly into the loader so that they're no longer required.

mikebattista (approved):

mikebattista (approved):

PR #1272, opened 2022-09-22
Windows.Win32.System.RemoteDesktop.IWTSVirtualChannelManager.CreateListener : pszChannelName should be a null terminated string instead of Byte*

### LINKED ISSUE #1185: IWTSVirtualChannelManager.CreateListener expects a string of type &u8, which might be a regression
microsoft/windows-rs#1939 changed IWTSVirtualChannelManager.CreateListener to take a &u8 instead of a *const u8. I was able to get a *const u8 by simply calling as_ptr() on a null terminated string, but obviously that does no longer work. I can prepend &* and that works, but it is not ideal. &u8 would suggest a reference to a single u8 value rather than to a null terminated string.

_Originally posted by @leonardder in https://github.com/microsoft/windows-rs/issues/1939#issuecomment-1228746322_

leonardder:
Fixes #1185 

Note, I'm slightly puzzled because the parameter is const char* in the source, therefore I'd expected this being emitted as PSTR already without this explicit intervention.


mikebattista:
@sotteson1 / @chenss3 could you please review to understand if the tooling is behaving as expected here before we approve the manual override?

sotteson1:
> @sotteson1 / @chenss3 could you please review to understand if the tooling is behaving as expected here before we approve the manual override?

The problem is in the original definition, it explicitly uses const unsigned char * instead of a type we recognize as being a string (like PSTR or LPCSTR). So, it's expected that we would need to add more data to the emitter.

leonardder:
Should I close this pr and create a new one that patches the emitter to cover these cases?

leonardder:
I had a quick look at the tsvirtualchannels.h header in generation/WinSDK/RecompiledIdlHeaders/um and indeed, there the definition is `_In_  const unsigned char *pszChannelName,`. However, in tsvirtualchannels.idl, the definition is `[in, string]    const char          *pszChannelName,`. Also in `tsvirtualchannels.h` in windows SDK 10.0.22621.0, the unsignedness is missing.

mikebattista:
Is there an infrastructure issue here or is "patching the emitter" just what you originally did by updating the emitter settings?

leonardder:
I tried updating the emitter by adding "const unsigned char *" to the several cases that emit a pstr, but that didn't changeanything, interestingly.

riverar:
> [...] So, it's expected that we would need to add more data to the emitter.
I believe @sotteson1's point was that it is **expected** to need to feed a more precise type in as this PR is doing. I don't believe he was suggesting that anyone go try to modify the emitter.

That said, we could, as part of a separate task, do something with the `__RPC__in_string` annotation.

riverar (approved):

BenJKuhn:
Sorry @sotteson1, I seem to have accidentally removed you from reviewers somehow when adding & removing myself to test something, and github won't let me add you back. Feel free to add yourself again.

leonardder:
I'm slightly puzzled, when I merge main into this branch, I don't get a merge conflict for scripts/ChangesSinceLastRelease.txt and somehow unrelated lines end up in the diff. So I reverted the merge of main for now.

leonardder:
Ah, it instantly became obvious to me that ChangesSinceLastRelease.txt was entirely rebuild since we had a release since my initial commit.

mikebattista (approved):

PR #1320, opened 2022-10-19
Added VARIANT_BOOL

### LINKED ISSUE #1319: VARIANT_BOOL isn't in the metdata
WinForms uses `VARIANT_BOOL` and `VARIANT_TRUE` and `VARIANT_FALSE`. It is defined in `wtypes.h` and used with `VARIANT` and various OLE interfaces.

mikebattista:
Fixed #1319.

VARIANT_BOOL is used across the API surface so I added the enum itself to Windows.Win32.Foundation. Then I remapped all occurrences of VARIANT_BOOL in the headers to use the new enum via the scraper settings file.


elachlan:
Values in the enum seem correct. Lots of API changes, but it looks correct to me.

chenss3 (approved):

kennykerr:
Seems reasonable.

PR #1321, opened 2022-10-21
Replaced PreserveSig usage with explicit attributes

### LINKED ISSUE #1315: Annotate members that return failing HRESULTs in success paths
When APIs are declared in the metadata *without* `preservesig`, some projections (e.g. CsWin32) will produce an API that *must* throw an exception when a failing HRESULT is returned. This actually provides a good C# coding experience *most* of the time. But when an API *regularly* returns failing HRESULTs, even in 'success' paths (e.g. reaching the end of a list), this raises runtime cost and coding complexity due to the need to throw and handle exceptions where a simple return value check would have been preferable.

In #1295, we learned that Rust has a counter-interest where they want to see `preservesig` *only* for methods that return non-S_OK success codes. They want the metadata to _omit_ `preservesig` even if a method returns a failure code regularly. This makes sense, since rust has no exceptions.

We need to define an annotation for the metadata to apply to methods that return error codes in success paths so that it doesn't disrupt rust, but gives CsWin32 the hint it needs to apply `PreserveSig` semantics to the C# declaration. 

In the meantime, this issue will aggregate the list of all APIs that we intend to apply this annotation to when it is invented.

APIs follow:
- DXGIFactory::EnumAdapters
- IDXGIFactory1::EnumAdapters1
- IDXGIAdapter::EnumOutputs
- ID2D1RenderTarget::EndDraw

If you have an API to add to this list, edit this description to add it, or comment below.

Also note that CsWin32 has a user-applicable workaround for any of these APIs that can unblock them. The NativeMethods.json file can include a preservesig property where you can list any particular API that you want declared with PreserveSig:

```json
  "comInterop": {
    "preserveSigMethods": [ "ID2D1DCRenderTarget.EndDraw" ]
  }
```

mikebattista:
Fixed #1315.


kennykerr (approved):

riverar (approved):
Looks good to me.

Is everyone happy with those names? Naming things is hard, now is a good time to scrutinize the names before it goes out.

AArnott (approved):

mikebattista:
Yes I'm happy with the names.

PR #1347, opened 2022-10-31
Change return type of some functions in POWRPROF.dll to WIN32_ERROR

### LINKED ISSUE #1346: The return type of the POWRPROF API should be WIN32_ERROR
Functions in PowrProf.dll (such as [PowerEnumerate](https://learn.microsoft.com/en-us/windows/win32/api/powrprof/nf-powrprof-powerenumerate)), returns Win32 error codes.
Since these error codes are represented in the metadata as `WIN32_ERROR`, the return type of these functions should be `WIN32_ERROR`, not `uint`.

aetos382:
fix #1346


aetos382:
Close to change the source branch of a pull request.

PR #1380, opened 2022-11-22
Sign the Microsoft.Windows.Sdk.Win32Docs.dll assembly

AArnott:
We need to build it, sign it, then pack it.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -33,6 +33,6 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Nerdbank.GitVersioning" Version="3.4.231" PrivateAssets="all" />
+    <PackageReference Include="Nerdbank.GitVersioning" Version="3.5.119" PrivateAssets="all" />
   </ItemGroup>
 </Project>
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -343,7 +343,52 @@ jobs:
       version: 2.1.x
   - powershell: dotnet tool update --global nbgv
     displayName: ⚙ Install nbgv
-  - script: dotnet pack -c $(BuildConfiguration)
+  - script: dotnet build -c $(BuildConfiguration)
+    displayName: 🏭 dotnet build
+    workingDirectory: apidocs
+  - task: EsrpCodeSigning@1
+    displayName: ✒ Assembly sign
+    inputs:
+      ConnectedServiceName: Xlang Code Signing
+      FolderPath: $(System.DefaultWorkingDirectory)/bin/$(BuildConfiguration)/netstandard2.0
+      Pattern: '*.dll'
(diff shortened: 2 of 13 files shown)


AArnott:
I've verified that the dll is now signed, and that the docs package gets a new version so that it will actually push.

mikebattista (approved):

riverar:
What's the new assembly for?

AArnott:
What new assembly? This change doesn't introduce a new assembly. It just signs an existing one.
If you're wondering what Microsoft.Windows.Sdk.Win32docs.dll is for, it just defines the schema for the apidocs.msgpack file.

riverar:
Ah neat, thanks. Wasn't familiar with this assembly.

timsneath:
@AArnott is it documented somewhere by any chance?

timsneath:
Is it used by another projection like C#? If there's some existing code that could act as a breadcrumb trail for me to follow, I'd appreciate anything you can offer!

AArnott:
@timsneath, breadcrumb follows:
1. note [the nuget dependency from my projection to the docs package](https://www.nuget.org/packages/microsoft.windows.cswin32#dependencies-body-tab). See also [source](https://github.com/microsoft/CsWin32/blob/78b48564d09b671eb707862c416e990609c5ff74/src/Microsoft.Windows.CsWin32/Microsoft.Windows.CsWin32.nuspec#L20). This package contains both the assembly that defines the schema and the msgpack docs content itself.
2. [How I deserialize the docs file](https://github.com/microsoft/CsWin32/blob/main/src/Microsoft.Windows.CsWin32/Docs.cs). IIRC `ApiDetails` is the type defined within the Microsoft.Windows.Sdk.Win32Docs.dll assembly. I *find* the file to deserialize [via an msbuild property](https://github.com/microsoft/CsWin32/blob/78b48564d09b671eb707862c416e990609c5ff74/src/Microsoft.Windows.CsWin32/SourceGenerator.cs#L378-L404) that [is set here](https://github.com/microsoft/CsWin32/blob/78b48564d09b671eb707862c416e990609c5ff74/src/Microsoft.Windows.CsWin32/build/Microsoft.Windows.CsWin32.props#L21) based on [an msbuild item that is defined by the docs package itself](https://github.com/microsoft/win32metadata/blob/d3ecea2d1d6ee3ca757569d6bfa2fb955b87fb6b/apidocs/Microsoft.Windows.SDK.Win32Docs/buildTransitive/Microsoft.Windows.SDK.Win32Docs.props#L3).
3. [How I actually produce the source doc comments in the emitted code](https://github.com/microsoft/CsWin32/blob/main/src/Microsoft.Windows.CsWin32/Generator.ApiDocs.cs).

timsneath:
Thanks Andrew! I'll take a look.

Issue #1330, opened 2022-10-26
Validate struct sizes against MSVC


mikebattista:
Since this isn't specific to Rust, it would be great if you have some kind of build validation that compared all struct sizes in metadata against MSVC.

_Originally posted by @kennykerr in https://github.com/microsoft/win32metadata/issues/1325#issuecomment-1290624646_

mikebattista:
Since the only known failures here have been misapplied enums, we should consider extending the existing validation we have for methods to look at structs as well.

kennykerr:
As a short-term solution, perhaps ClangSharp will give you the size of the original struct?

mikebattista:
@chenss3 any updates on this?

PR #1400, opened 2022-12-04
Add CONTEXT_* values to enums.json

### LINKED ISSUE #556: Missing type / consts for CONTEXT.ContextFlags
bindings::Windows::Win32::System::Diagnostics::Debug::CONTEXT as a field named ContextFlags.

The values that one may set for this are difficult to find though they seem to be documented here: https://github.com/tpn/winsdk-10/blob/master/Include/10.0.14393.0/km/ntddk.h#L493-L517

There should either be a struct for this field, or at least just some consts for setting it properly somewhere in the metadata

### LINKED ISSUE #1042: Feature request: No impelemtation CONTEXT_CONTROL, CONTEXT_FULL etc for windows::Win32::System::Diagnostics::Debug::CONTEXT

CraftSpider:
Fixes microsoft/wdkmetadata#5 
Fixes microsoft/win32metadata#556

The scraper isn't smart enough to handle these, and some of the values are different based on architecture. I went with defining the constants without the current platform implicitly included. It may be a good idea to also define all of them with an architecture prefix - only have `CONTEXT_AMD64_*` and such, not including `CONTEXT_INTEGER` and such directly.


CraftSpider:
@microsoft-github-policy-service agree

mikebattista:
Based on your comments, could you clarify any remaining spec issues that are left here? Does the current change work for all architectures as is?

CraftSpider:
Currently, `CONTEXT_*` covers all of x86_64 and `CONTEXT_AMD64_*` covers both `arm` and `arm64`. There are two issues:
- Using `CONTEXT_*` on arm64 would get you the wrong values but not be obvious, so one question is whether to move those under `CONTEXT_AMD64_*`
- `CONTEXT_ARM64_ALL` includes `CONTEXT_ARM64_X18` which isn't defined on `ARM`, so the second question is whether to mostly duplicate it into `CONTEXT_ARM_*` and `CONTEXT_ARM64_*`
- All architectures must always or in their platform constant (EG `CONTEXT_FULL | CONTEXT_AMD64`) to avoid accidentally including the wrong architecture value.

The current changes does allow correct code to be written on all architectures, the issues are simply about whether there is a way to make it *easier* to write that correct code.

riverar:
`CONTEXT_CONTROL` is...

* `0x000001` in win32metadata / this PR
* `0x010001` on x86
* `0x100001` on x64/amd64
* `0x200001` on arm
* `0x400001` on arm64

I don't think we can move forward with this as-is. Instead, perhaps we should add an "architecture" key to `enums.json` so that we can emit the constants with the appropriate `[SupportedArchitecture(...)]` attribute?

Example:
```json
{
    "name": "CONTEXT_FLAGS",
    "type": "uint",
    "flags": true,
    "architecture": "x86",
    "members": [
      ...
      {
        "name": "CONTEXT_CONTROL",
        "value": "0x010001"
      },
```

Would emit:
```csharp
[SupportedArchitecture(Windows.Win32.Interop.Architecture.X86)]
enum CONTEXT_FLAGS { ... }
```

riverar (changes requested):
`CONTEXT_CONTROL` is...

* `0x000001` in win32metadata / this PR
* `0x010001` on x86
* `0x100001` on x64/amd64
* `0x200001` on arm
* `0x400001` on arm64

I don't think we can move forward with this as-is. Instead, perhaps we should add support for an "architecture" key to `enums.json` so that we can emit the constants with the appropriate `[SupportedArchitecture(...)]` attribute?

Example:
```patch
{
  "name": "CONTEXT_FLAGS",
  "type": "uint",
  "flags": true,
+ "architecture": "x86",
  "members": [
    // ...
    {
      "name": "CONTEXT_CONTROL",
      "value": "0x010001"
    },
    // ...
}
```

Would emit:
```csharp
[SupportedArchitecture(Windows.Win32.Interop.Architecture.X86)]
enum CONTEXT_FLAGS : uint {
  // ...
  CONTEXT_CONTROL = 0x010001,
  // ...
}
```

Bee7736 (approved):

CraftSpider:
I can try to look into the architecture thing, but I'm not totally confident in my familiarity with this codebase to do that in a timely fashion. But I'll do that, and separately consider a more consistent handling of the constant values.

riverar:
@CraftSpider May want to wait until we get some buy in from @mikebattista. Then we need to verify this is something projection folks downstream can handle without blowing up, etc.

kennykerr:
Seems like we'd need more rigorous validation (e.g. #1330) before we entertain changes like this.

mikebattista:
ARM64 constants are qualified like below.

```
#define CONTEXT_ARM64   0x00400000L

#define CONTEXT_ARM64_CONTROL (CONTEXT_ARM64 | 0x1L)
#define CONTEXT_ARM64_INTEGER (CONTEXT_ARM64 | 0x2L)
#define CONTEXT_ARM64_FLOATING_POINT  (CONTEXT_ARM64 | 0x4L)
#define CONTEXT_ARM64_DEBUG_REGISTERS (CONTEXT_ARM64 | 0x8L)
#define CONTEXT_ARM64_X18 (CONTEXT_ARM64 | 0x10L)
```

What if we just have one enum that has all the values for all architectures and we quality each name with the architecture like above.

* `CONTEXT_AMD64_CONTROL`
* `CONTEXT_ARM64_CONTROL`
* `CONTEXT_i386_CONTROL`
* `etc.`

Or we could move the prefix to the beginning/end so the original constant name (`CONTEXT_CONTROL`) isn't disrupted.

* `CONTEXT_CONTROL_AMD64`
* `CONTEXT_CONTROL_ARM64`
* `CONTEXT_CONTROL_i386`
* `etc.`

CraftSpider:
Thanks for updating the PR, sorry for not responding - the always-suffixed constants look good to me, and I think were what I was trying to get at in my opening comment about 'only having `CONTEXT_AMD64_*`. I definitely prefer the architecture at the end style.

mikebattista (approved):

riverar (commented):
I don't like the fake context structures added to metadata. We should look into fixing this properly asap before projection authors start generating new code.

mikebattista:
We can remove the constants altogether or unassign them from the APIs so they can be safely ignored, but implementing per-architecture enums is not something that will be fixed quickly.

mikebattista:
Right I'd definitely like to at least get the constants in there than not have them.

Could you clarify how you use this struct and flags in a multi-architecture project? Do you typically have shared code and trust that the flags will update when you switch the target? Or do you already have architecture-specific code paths where you could easily switch to the fully qualified constant names from the enum?

CraftSpider:
I'm in Rust, and can't speak to other languages or projects, but currently I have things subdivided by OS (I'm trying to support multiple) then by architecture, for Windows I define a helper for the flags with architecture-specific config code internally (though I rely on the CONTEXT struct being correct per-arch) - I can personally drop in hardware specific flag names easily enough, and thinking about it, I might not always be able to use shared code anyways - since some archs have different flags available, and `_ALL` may not always be what I want.

PR #1405, opened 2022-12-09
Fixed SAL parsing of _opt

### LINKED ISSUE #1005: `_Outptr_result_maybenull_` and variants interpreted incorrectly
Like a few previous issues, this SAL annotation denotes that the result may be null, not that the parameter itself is optional. For example, `ID2D1DeviceContext::GetTarget` is described as follows:

```C#
unsafe void GetTarget([Optional][Out] ID2D1Image* image);
```

But this parameter is not in fact optional. There are also plenty of other parameters with these annotations being interpreted incorrectly. These include:

* `_Outptr_result_maybenull_`
* `_Outptr_result_buffer_maybenull_`
* `_COM_Outptr_result_maybenull_`

There may well be others.

mikebattista:
Fixed #1005.


mikebattista:
@riverara @kennykerr please review the changes and diff. Many Optional attributes were removed including the ones mentioned in #1005. I spot checked several and they seem to be expected changes. It's possible there are regressions here or patterns that aren't scanned as they should.

There was one edge case for a handful of APIs like SendMessageA/W that use _Pre_maybenull and _Post_valid instead of the _In_ _Opt_ pattern. For those, I just check for _Pre_maybenull to indicate [In][Optional]. If I understand the semantics correctly and how we were parsing them, they seem to be conflicting so I prioritized _Pre_maybenull if that is declared.

scripts/ChangesSinceLastRelease.txt
```diff
@@ -236,3 +236,296 @@ Windows.Win32.System.Threading.Apis.CreateProcessAsUserW : dwCreationFlags...UIn
...
+Windows.Win32.System.Com.Urlmon.IInternetHostSecurityManager.QueryCustomPolicy : ppPolicy : [NativeArrayInfo(CountParamIndex=2),Optional,Out] => [NativeArrayInfo(CountParamIndex=2),Out]
+Windows.Win32.System.Diagnostics.Debug.Apis.MiniDumpReadDumpStream : Dir : [Optional,Out] => [Out]
+Windows.Win32.System.Diagnostics.Debug.Apis.MiniDumpReadDumpStream : StreamPointer : [Optional,Out] => [Out]
+Windows.Win32.System.Diagnostics.Debug.IDynamicConceptProviderConcept.GetConcept : conceptInterface : [ComOutPtr,Optional,Out] => [ComOutPtr,Out]
+Windows.Win32.System.Diagnostics.Debug.IModelObject.GetContext : context : [ComOutPtr,Optional,Out] => [ComOutPtr,Out]
+Windows.Win32.System.Diagnostics.Debug.IModelObject.GetParentModel : contextObject : [ComOutPtr,Optional,Out] => [ComOutPtr,Out]
+Windows.Win32.System.Diagnostics.Etw.Apis.TdhCreatePayloadFilter : PayloadFilter : [Optional,Out] => [Out]
+Windows.Win32.System.EventLog.Apis.EvtSubscribe : Callback : [Optional] => [In]
+Windows.Win32.System.EventLog.Apis.EvtSubscribe : Context : [Optional] => [In,Out]
+Windows.Win32.System.EventLog.EVT_SUBSCRIBE_CALLBACK.Invoke : UserContext : [Optional] => [In,Out]
```
riverar, line 476: Callback is marked as `_Maybenull_` in the headers. This can be set to `null` or point to a blob somewhere. This seems to qualify for the `[Optional]` attribute, right?

mikebattista: We can check for _maybenull_ by itself and add optional for cases like this.

scripts/ChangesSinceLastRelease.txt
```diff
@@ -236,3 +236,296 @@ Windows.Win32.System.Threading.Apis.CreateProcessAsUserW : dwCreationFlags...UIn
...
+Windows.Win32.System.Com.IUriBuilder.GetUserName : ppwzUserName : [Optional,Out] => [Out]
+Windows.Win32.System.Com.Urlmon.IInternetHostSecurityManager.QueryCustomPolicy : ppPolicy : [NativeArrayInfo(CountParamIndex=2),Optional,Out] => [NativeArrayInfo(CountParamIndex=2),Out]
+Windows.Win32.System.Diagnostics.Debug.Apis.MiniDumpReadDumpStream : Dir : [Optional,Out] => [Out]
+Windows.Win32.System.Diagnostics.Debug.Apis.MiniDumpReadDumpStream : StreamPointer : [Optional,Out] => [Out]
+Windows.Win32.System.Diagnostics.Debug.IDynamicConceptProviderConcept.GetConcept : conceptInterface : [ComOutPtr,Optional,Out] => [ComOutPtr,Out]
+Windows.Win32.System.Diagnostics.Debug.IModelObject.GetContext : context : [ComOutPtr,Optional,Out] => [ComOutPtr,Out]
+Windows.Win32.System.Diagnostics.Debug.IModelObject.GetParentModel : contextObject : [ComOutPtr,Optional,Out] => [ComOutPtr,Out]
+Windows.Win32.System.Diagnostics.Etw.Apis.TdhCreatePayloadFilter : PayloadFilter : [Optional,Out] => [Out]
+Windows.Win32.System.EventLog.Apis.EvtSubscribe : Callback : [Optional] => [In]
+Windows.Win32.System.EventLog.Apis.EvtSubscribe : Context : [Optional] => [In,Out]
```
riverar: `_Maybenull_ PVOID Context`

I don't see anything in the headers or docs that indicate this is an output.

mikebattista: Fixed with the latest change. Will be [In][Optional].

Issue #1426, opened 2023-01-12
Generating metadata for `IInspectable`-derived interfaces is broken


kennykerr:
For example, the following IDL will cause the generator to produce invalid interface definitions with duplicate vtable entries:

```C++
[
    object,
    uuid(7dc67935-9dfc-4129-a33a-3784b4250569)
]
interface IFoo : IInspectable
{
    HRESULT SetFoo();
}
```

riverar:
Steps to reproduce

In a developer command prompt, ...

1. `git clone https://github.com/riverar/repro_2d66609dd405`
2. `cd repro_2d66609dd405`
3. `dotnet build`
4. Inspect `Output.winmd` with ILSpy or other tooling
5. Observe incorrect interface methods were emitted

```csharp
[Guid(2980787302u, 29309, 19891, 150, 162, 23, 97, 60, 61, 39, 34)]
public interface ISimple : IInspectable
{
	new unsafe HRESULT GetIids([Out] uint* iidCount, [Out][NativeArrayInfo(CountParamIndex = 0)] Guid** iids);
	new unsafe HRESULT GetRuntimeClassName([Out] HSTRING* className);
	new unsafe HRESULT GetTrustLevel([Out] TrustLevel* trustLevel);
	HRESULT DoSomethingSimple();
}
```

PR #1427, opened 2023-01-13
Fix method counting for interfaces when scanning references

riverar:
`ClangSharpSourceWinmdGenerator` maintains a cache of encountered interface types and their method counts. This information is currently obtained during two phases:
1. a walk of input syntax trees
2. a walk of compilation references

The input syntax trees (1) contain interface types with all inherited and local methods already emitted. When `ClangSharpSourceWinmdGenerator` performs a `MethodDeclarationSyntax` count, it caches the correct value.

Example input fragment:
```csharp
[NativeInheritance("IInspectable")]
public unsafe partial struct IDragDropManagerInterop
{
  public delegate int _QueryInterface(IDragDropManagerInterop* pThis, [NativeTypeName("const IID &"),Const]Guid* riid, [ComOutPtr]void** ppvObject);
  public delegate uint _AddRef(IDragDropManagerInterop* pThis);
  public delegate uint _Release(IDragDropManagerInterop* pThis);
  ...
}
```

The compilation references (2) (e.g. Windows.Win32.winmd) however contain interface types that do not have these methods emitted. `ClangSharpSourceWinmdGenerator` instead counts the number of methods on the type specified by `NativeInheritance` and moves on, resulting in an inaccurate count. This count is used later to skip over inherited methods when emitting interface method metadata, resulting in interfaces with invalid method members.

Example:
```csharp
public interface IFoo : IInspectable
{
  // 3 IUnknown methods skipped over

  // 3 IInspectable methods that didn't get skipped over
  unsafe HRESULT GetIids(...);
  unsafe HRESULT GetRuntimeClassName(...);
  unsafe HRESULT GetTrustLevel(...);
  unsafe HRESULT IFooFirstMethod(...);
}
```

This impacts all projects with references to Windows.Win32.winmd and using the Microsoft.Windows.WinmdGenerator to generate metadata. (Generally everyone outside of win32metadata.)

This PR fixes method counting during the compilation reference walk by also counting the methods for each interface implemented by an interface type. This information is obtained via the `InterfaceImpl` table (`II.22.23`).

Fixes: #1426

### DIFF
--- a/sources/ClangSharpSourceToWinmd/ClangSharpSourceWinmdGenerator.cs
+++ b/sources/ClangSharpSourceToWinmd/ClangSharpSourceWinmdGenerator.cs
@@ -429,7 +429,7 @@ namespace ClangSharpSourceToWinmd
             if (symbol != null)
             {
                 this.interfaceSymbols.Add(symbol);
-                this.interfaceMethodCount[symbol] = interfaceInfo.Methods.Count();
+                this.interfaceMethodCount[symbol] = interfaceInfo.ImplementedMethodCount;
                 if (!this.namesToInterfaceSymbols.TryGetValue(symbol.Name, out var symbols))
                 {
                     symbols = new List<ISymbol>();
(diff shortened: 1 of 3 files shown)


BenJKuhn:
Change looks great, thanks for including new test collateral, and for tackling this issue!

BenJKuhn (approved):

mikebattista (approved):

Issue #1428, opened 2023-01-13
Make it possible to customize MIDL parameters


abolade:
The problem: MetadataTasks.dll currently uses hard-coded parameters for MIDL compilation. However, certain IDL files require that /notlb be specified as a parameter (e.g. those using system_handle(...) parameters on interface methods), as otherwise midl.exe fails with an error.

The proposed solution: Provide a build directive that allows MIDL parameters to be customized in the CompileIdls task, e.g.:

```
namespace MetadataTasks
{
    public class CompileIdls : Task
    {
< snip >
        public string MidlSwitches
        {
            get; set;
        }
```

BenJKuhn:
I only have a cursory understanding of TLBs. For our purposes, is there any reason we can't just always pass /notlb? Are there IDLs that require that we generate them? Allowing customization seems like a reasonable plan either way.

mikebattista:
See https://github.com/microsoft/win32metadata/pull/1432#issuecomment-1396230501 for PR and comment on changing the default.

PR #1432, opened 2023-01-18
Added MidlSwitches for #1428.

mikebattista:
I followed the existing pattern for AdditionalIncludes to add MidlSwitches for #1428.


mikebattista:
@benkuhn if we wanted to set /notlb as the default, I believe we would make that change in `commonMidlArgs`.

BenJKuhn (approved):

abolade (approved):
This looks fine for simply appending switches to the midl.exe command line.

If the command line needs to be replaced, a different approach might be needed: check if MidlSwitches is null or empty and, if so, use the defaults currently set in commonMidlArgs; otherwise, use the contents of MidlSwitches *instead of* the defaults currently set in commonMidlArgs.

mikebattista:
Thanks. For now, I'd like to start with being able to append flags that are missing.

Unless we want to make a change to set /notld by default per @benkuhn's comments, then we can set that switch in the default arguments.

Are there other MIDL parameters that would need to be appended in this way besides /notld? If we made /notld the default, would we still want to have this ability to append additional arguments?

chenss3 (approved):

mikebattista:
If we want to keep this, I believe the below would also need to be updated to pass through the user-defined switches to the CompileIdls Task.

https://github.com/microsoft/win32metadata/blob/e6e1456e6ca0251d6d23134b80b242280bb841be/sources/GeneratorSdk/sdk/sdk.targets#L124-L133

BenJKuhn (dismissed):

mikebattista:
What are the implications of setting /notld by default? Is that expected to work in all scenarios?

kennykerr:
That should be safe. Since you aren't consuming tlb files, you can safely avoid generating them.

mikebattista:
Ok. I will set it by default and leave the custom additional flags support in. It's optional so there's no harm in leaving it there as an experiment.

PR #1445, opened 2023-01-25
Override D3D headers with those from D3D12 "Agility SDK" nuget package

### LINKED ISSUE #1437: Use latest DirectX-Headers
The latest version of many of the D3D headers can be found at https://github.com/microsoft/DirectX-Headers, and it would be great if win32metadata could use these rather than the ones currently found in https://github.com/microsoft/win32metadata/tree/main/generation/WinSDK/RecompiledIdlHeaders.

It's not obvious what the right approach is for this, so opening this issue for discussion before anyone starts on a PR.  Also this would be a good opportunity to validate that this is the right thing to do.

Some options:

1. Use a submodule?
2. Manually submit a PR to update the headers under RecompiledIdlHeader.  How would this work the next time the WinSDK is updated?
3. Manually submit a PR to update the headers under some other location.

Maybe there's some other approaches that might work?

damyanp:
Direct3D releases new D3D features via the [Agility SDK](https://devblogs.microsoft.com/directx/gettingstarted-dx12agility/) at a more regular cadence than OS releases.  These new features all eventually make it into the Windows OS (and therefore the Windows SDK) but developers often need to build against the Agility SDK.  This change makes it so that `RecompileIdlFilesForScraping.ps1` overwrites the D3D headers with the latest version from the Microsoft.Direct3D.D3D12 nuget package.

The stable APIs in the Agility SDK are always compatible with binaries that ship with the OS (normal COM interface discovery enables this).  Developers who wish to use an API that isn't yet available across all builds of Windows will also need to obtain the binaries contained in the nuget package and ship them alongside their application (see the link above for details).

Resolves #1437.


scripts/RecompileIdlFilesForScraping.ps1
```diff
@@ -22,6 +22,12 @@ copy-item -Path "$sdkIncludeDir\shared" -destination "$recompiledIdlHeadersDir"
 copy-item -Path "$sdkIncludeDir\winrt" -destination "$recompiledIdlHeadersDir" -recurse
 copy-item -Path "$sdkIncludeDir\ucrt" -destination "$recompiledIdlHeadersDir" -recurse
 
+$d3dPkgPath = Get-BuildToolsNugetPropsProperty("PkgMicrosoft_Direct3D_D3D12")
+$d3dIncludeDir = Join-Path $d3dPkgPath "build/native/include/"
+Write-Host "Updating to latest Direct3D headers...$d3dIncludeDir to $recompiledIdlHeadersDir"
+copy-item -Path "$d3dIncludeDir\*.*" -destination "$recompiledIdlHeadersDir\um" -Exclude dxgiformat.* -recurse
+copy-item -Path "$d3dIncludeDir\dxgiformat.*" -destination "$recompiledIdlHeadersDir\shared" -Exclude dxgiformat.* -recurse
```
jenatali, line 29: The `-Exclude dxgiformat.*` here seems like a copy/paste error, considering that's the wildcard you're *trying* to copy.

jenatali (approved):
One comment but otherwise LGTM.

scripts/RecompileIdlFilesForScraping.ps1
damyanp: Oops, nice catch!  Thanks!

mikebattista (approved):

mikebattista:
Can you review the diff at https://github.com/microsoft/win32metadata/commit/72b63a640d37490c57d4e8db0f664d4860ba2489 to make sure everything is expected?

I see at least 1 removal, and some changed values near the end.

jenatali:
Yep, all of those changes look good to me.

kennykerr:
I can update windows-rs as soon as a new release is published.

mikebattista:
> I can update windows-rs as soon as a new release is published.

I'm publishing an update now.

MarijnS95:
@mikebattista thanks! @kennykerr can you import and regenerate bindings with the new win32metadata drop? That'll allow us to test this out straight from a `git` dependency on the `windows` crate.

kennykerr:
https://github.com/microsoft/windows-rs/pull/2322

MarijnS95:
Thanks!

damyanp:
When we release a new Agility SDK we update the win32metadata project with the latest headers, and a short time after that Kenny usually releases a new windows-rs that ingests this new metadata, so you shouldn't need to do anything to teach windows-rs about any new APIs in the Agility SDK.

You'll need to follow the "Set Agility SDK Parameters" section described in that blog, and make sure that the appropriate Agility SDK DLLs are available for your application.

I don't know if anyone has tried this before though.  I've filed an issue suggesting we add a sample for this. (https://github.com/microsoft/DirectX-Graphics-Samples/issues/866)

MarijnS95:
@matteomeli "it depends". I'd say the Agility SDK consists of two parts: more up to date headers defining existing and newer (WIP?) Direct3D12 bindings, and a binary part that _implements_ said (WIP?) bindings.

For this particular change in `win32metadata` and `windows-rs`, the bindings can be used as-is, and feature queries work as normal via COM's `QueryInterface` and the like, as if you were using Direct3D12 bindings before this change.

But in order to take advantage of anything new that the Agility SDK has to offer, as you request, you have to translate the linked guide to Rust. It took us some tinkering (and I'm not sure if there's any example with this up online) to get the exports to work, but all you effectively need to do to export the symbols is:

```rust
#[no_mangle]
pub static D3D12SDKVersion: u32 = 610;
#[no_mangle]
pub static D3D12SDKPath: &[u8; 18] = b"dx12-agility-sdk\\\0";
```

(Adjust values per your need. `610` is the version, `dx12-agility-sdk\` is the path to a folder next to our `.exe` containing the downloaded `.dll`s)
(I see now that we might even replace this with a `CStr`, and taking it a step further: use the new Rust 1.77 c-string literals: `pub static D3D12SDKPath: &std::ffi::CStr = c"dx12-agility-sdk\\";`)

Since symbols won't get exported from an exe by default, we have to set up an export file: in a `build.rs` file you specify it as follows:

```rust
println!(
    r#"cargo:rustc-link-arg-bins=/DEF:{}"#,
    exports_path.display()
);
```

And this file pointed to by `exports_path` (`exports.def`) contains the same thing as the upstream guide:

```exports
EXPORTS
  D3D12SDKVersion DATA PRIVATE
  D3D12SDKPath DATA PRIVATE
```

Good luck!

matteomeli:
That sounds good. Enough information for me to give it a go. 

How would I know that the binding of an Agility SDK version has happened successfully? IS there a way to verify it?

jenatali:
I'll also add that the "independent devices" feature documented [here](https://microsoft.github.io/DirectX-Specs/d3d/IndependentDevices.html#api) has been released (note to self, update the spec to clarify that it's not a proposal anymore). This feature is available on up-to-date Windows 10 and 11, and allows configuring the Agility SDK at runtime instead of using static executable metadata.

> How would I know that the binding of an Agility SDK version has happened successfully? Is there a way to verify that it worked aside the program compiles and runs?
There's two options people typically use:
1. Use a debugger and check the path where D3D12Core.dll is loaded from.
2. Try to query for an interface which isn't available on the OS you're actually running. E.g. querying for `ID3D12Device14` will only succeed if you're using Agility SDK 1.613.0 or newer, since that interface is not available on any in-market OS today.

matteomeli:
Managed to get it working.

In the source for the exe somewhere:
```
#[cfg(target_os="windows")]
#[allow(non_upper_case_globals)]
#[no_mangle]
pub static D3D12SDKVersion: u32 = 613;
#[cfg(target_os="windows")]
#[allow(non_upper_case_globals)]
#[no_mangle]
pub static D3D12SDKPath: &std::ffi::CStr = c"D3D12\\";
```

In the `build.rs` file:
```
use std::env;
use std::path::{Path, PathBuf};

const DX12_AGILITY_SDK_LIB_PATH: &str = "extern\\dx12-agility-sdk";

fn main() {
    let workspace_dir = workspace_dir();

    // Export DX12 Agility SDK constants via a .def file
    println!(
        "cargo:rustc-link-arg=/DEF:{}\\exports.def",
        workspace_dir
            .join(DX12_AGILITY_SDK_LIB_PATH)
            .to_str()
            .unwrap()
    );

    // Copy DX12 Agility SDK libs
    let copy_source_path = workspace_dir.join(DX12_AGILITY_SDK_LIB_PATH);
    let profile = env::var("PROFILE").unwrap();
    let bin_path = workspace_dir.join("target").join(profile);
    let copy_dest_path = bin_path.join("D3D12");
    std::fs::create_dir_all(&copy_dest_path).unwrap_or_else(|err| {
        eprintln!(
            "Cannot create D3D12 dir to copy DX12 Agility SDK dlls: {:?}",
            err
        );
    });

    let files_to_copy = [
        "D3D12Core.dll",
        "D3D12Core.pdb",
        "d3d12SDKLayers.dll",
        "d3d12SDKLayers.pdb",
    ];

    for file in files_to_copy {
        std::fs::copy(copy_source_path.join(file), copy_dest_path.join(file)).unwrap_or_else(
            |err| {
                eprintln!("Cannot copy DX12 Agility SDK libs: {:?}", err);
                0
            },
        );
    }
}
```

When there something wrong in the configuration I get this error: `The D3D12 SDK version configuration of the host exe is invalid. (0x887E0003)`., when trying to create the D3D12 device This is if there is something wrong with the SDK DLLs, either they are missing or they are for a different version of the SDK than the one stated by the `D3D12SDKVersion` exported symbol.

When everything is working I get the following log printing:
```
let device = D3d12::create_device(adapter, d3d12::FeatureLevel::L11_0)?;
if let Ok(device13) = device.0.cast::<ID3D12Device13>() {
    log::info!("Agility SDK version {} loaded.", D3D12SDKVersion);
}
```

By the way, at the moment `windows-rs` does not have the `ID3D12Device14` yet, it's only up to `ID3D12Device13`.

MarijnS95:
@matteomeli glad to hear that you got this working so quickly, and with roughly the same `build.rs` script that I have here to copy the binaries to the output directory.

Seems like your example shows that exporting a `&CStr` has "roughly the same ABI" as `&[u8; N]` (though I wouldn't count on it, fat pointer for the slice length vs a "lean" pointer with just the array)

> By the way, at the moment `windows-rs` does not have the `ID3D12Device14` yet, it's only up to `ID3D12Device13`.

Correct. I (as an independent contributor) imported the latest `613` Agility SDK to `win32metadata` bindings in #1870 but there hasn't been a corresponding release yet (that would subsequently be imported to `windows-rs`).

If you need access to it _now_, I have forked branches on top of `master` and the latest `0.54` release (because there hasn't been a `windows 0.55` release to match `windows-core 0.54`...) at https://github.com/Traverse-Research/windows-rs/branches/all?query=613

Issue #1420, opened 2022-12-28
Bug: OUTLINETEXTMETRICW struct is incorrectly sized


psy-fidelious:
### Which crate is this about?

windows

### Crate version

0.43

### Summary

The struct size of OUTLINETEXTMETRICW  in C++ is 232 bytes instead of 264 which is what std::mem::size_of::<OUTLINETEXTMETRICW>() returns in rust. 
This seems to cause issues with GetOutlineTextMetricsW which fails in rust with a zero return value, when a reference to this struct with the corresponding size is passed to it. This doesn't seem to occur in C++.
Curiously, if you call GetOutlineTextMetricsW with a nullptr argument, you get a required size of 318 in both C++ and rust which is higher than the size of what both languages report for those structs. I suppose some alignment is done behind the scenes, but it's clear that the underlying code checks for this exact struct size as well.

### Toolchain version/configuration

windows rust 1.66

### Expected behavior

The struct sizes between C++ and rust should match

mikebattista:
Looks like another case of https://github.com/microsoft/win32metadata/issues/1330.

Issue #1325, opened 2022-10-22
DEVMODE structure has fields with incorrect types


Skrity:
### Which crate is this about?

windows

### Crate version

0.42.0

### Summary

When querying display via EnumDisplaySettings(A,W,ExA,ExW) display parameters get inserted into wrong fields. Tested on 2 pcs so far.

### Toolchain version/configuration

Default host: x86_64-pc-windows-gnu
rustup home:  C:\_scoop\persist\rustup\.rustup

stable-x86_64-pc-windows-gnu (default)
rustc 1.64.0 (a55dd71d5 2022-09-19)

### Reproducible example

```rust
fn main() {
    use windows::Win32::Graphics::Gdi::*;
    let mut dm = DEVMODEA::default();
    dm.dmSize = std::mem::size_of_val(&dm) as u16;
    let b = unsafe {EnumDisplaySettingsExA(windows::core::PCSTR::null(), ENUM_CURRENT_SETTINGS, &mut dm, 0u32) };
    println!("Result: {:?}",b);
    println!("dmBitsPerPel: {:?}", dm.dmBitsPerPel);
    println!("dmPelsWidth: {:?}", dm.dmPelsWidth);
    println!("dmPelsHeight: {:?}", dm.dmPelsHeight);
    println!("dmDisplayFrequency: {:?}", dm.dmDisplayFrequency);
    println!("others");
    println!("dmSpecVersion: {:?}", dm.dmSpecVersion);
    println!("dmDriverVersion: {:?}", dm.dmDriverVersion);
    println!("dmDriverExtra: {:?}", dm.dmDriverExtra);
    println!("dmLogPixels: {:?}", dm.dmLogPixels);
    println!("dmICMMethod: {:?}", dm.dmICMMethod);
}
```

### Crate manifest

```toml
[package]
name = "test_refresh"
version = "0.1.0"
edition = "2021"

# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html

[dependencies.windows]
version = "0.42.0"
features = ["Win32_Foundation", "Win32_Graphics_Gdi"]
```

### Expected behavior

Result: BOOL(1)
dmBitsPerPel: *
dmPelsWidth: 2560
dmPelsHeight: 1440
dmDisplayFrequency: 165
others
dmSpecVersion: *
dmDriverVersion: *
dmDriverExtra: *
dmLogPixels: *
dmICMMethod: *

### Actual behavior

Result: BOOL(1)
[...]

PR #1447, opened 2023-01-27
Fix Enum Size Validation Bug

chenss3:
Fixes: #1330
Fixes: #1420 

**Problem**: Originally, EnsureEnumSizeMatchesOriginalSize(), the method that is supposed to do the enum size validation, is not able to work as intended because the scraper did not emit a NativeTypeName, which is yields our problem since EnsureEnumSizeMatchesOriginalSize() assumes there will always be a NativeTypeName. 

**Solution**: I added code that will add a NativeTypeName for every field that doesn't have it, with the current type being the new NativeTypeName. 
This should be able to detect any incorrect enum sizes such as in #1325, where DEVMODE_COLOR, DEVMODE_TRUETYPE_OPTION, DEVMODE_COLLATE are defined as uint but should be ushort.


sources/ClangSharpSourceToWinmd/MetadataSyntaxTreeCleaner.cs
```diff
@@ -1114,6 +1116,38 @@ private bool GetRemapInfo(string fullName, out List<AttributeSyntax> listAttribu
...
+                            {
+                                var attrName = attr.Name.ToString();
+                                if (attrName == "NativeTypeName")
+                                {
+                                    hasNativeTypeName = true;
+                                }
+                            }
+                        }
+
+                        if (!hasNativeTypeName && !(currentType.Contains("IUnknown**")) && !(currentType.Contains("IInspectable**")))
```
chenss3, line 1135: The specific conditional check for the NativeTypeName IUnknown** and IInspectable** are for the only two specific occurrences where NativeTypeName gets added twice for its field. It's because the original NativeTypeName was not showing in the original attribute list, so this new NativeTypeName gets added, and then it becomes a duplicate. I'm not sure why this is happening, but for now I have this temporary fix.

mikebattista:
Based on the diff there was only 1 API that was flagged as incorrect and fixed. How much confidence do we have that the logic is working and we've caught all issues? @kennykerr mentioned he's heard from customers hitting issues like this pretty consistently so I'd expect there would have been more broken APIs. Have we fixed all the biggest offenders manually at this point?

chenss3:
> > Based on the diff there was only 1 API that was flagged as incorrect and fixed. How much confidence do we have that the logic is working and we've caught all issues? @kennykerr mentioned he's heard from customers hitting issues like this pretty consistently so I'd expect there would have been more broken APIs. Have we fixed all the biggest offenders manually at this point?
> 
> It seems unlikely that this has caught all the issues, but I can't be certain. #1330 is about validating far more than just field/enum replacements. If you check the size and alignment of structs in C++ you can quickly eliminate all kinds of size and packing issues and determine whether this is in fact accurate. Short of doing that, I'd suggest going through the various size and alignment issues that have been previously reported and testing whether this solution would have caught them.

I tested it with #1420, and my fix wasn't able to pick up on that bug. But the cause of the problem is still the same, it's not able to validate struct size because it's not scraping a NativeTypeName from those fields. I'm still looking into this now - I'm going to close and reopen this when it's ready again.

chenss3:
Reopened the PR because I realized the PAN_ issue #1420 were not being picked up because there was no corresponding size of type BYTE. After adding that, it was able to properly give an error that the enum is an incorrect size, and I fixed it accordingly.

Like Kenny suggested, I tested it out with other alignment issues that were reported. I tested it out with #1330, another struct enum issue, by removing the byte size and it was able to properly pick up that the enum was an incorrect size.

mikebattista (approved):

BenJKuhn (approved):

PR #1473, opened 2023-02-18
Correct [Show|Scroll]Window statuses, commands, and flags

### LINKED ISSUE #1460: SW_ Constants type-error
I was using the `windows_sys` crate, version "0.45.0".

While looking at the constants associated with the Win32 `ShowWindow` function, I noticed what seems to be an error in the type annotations.

All the `ShowWindow` constants (such as `SW_HIDE`) start with the `SW_` prefix and have the type `SHOW_WINDOW_CMD`.

However, there other constants (such as `SW_ERASE`) which have nothing to do with `ShowWindow`, and instead are associated with `ScrollWindowEx`. These constants still have the type `SHOW_WINDOW_CMD`, which seems to be an error.

There are also constants (such as `SW_PARENTCLOSING`) which, while related to `ShowWindow`, are not commands to be passed to the `ShowWindow` function as arguments, but possible values received in the `lParam` of the `WM_SHOWWINDOW` message. Whether or not it is an error that these constants also have the type `SHOW_WINDOW_CMD`, I am unsure of. 

Here are some links to the official documentation:
`ShowWindow`: https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-showwindow
`ScrollWindowEx`: https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-scrollwindowex
`WM_SHOWWINDOW`: https://learn.microsoft.com/en-us/windows/win32/winmsg/wm-showwindow
`SW_HIDE`: https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/UI/WindowsAndMessaging/constant.SW_HIDE.html
`SW_ERASE`: https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/UI/WindowsAndMessaging/constant.SW_ERASE.html
`SW_PARENTCLOSING`: https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/UI/WindowsAndMessaging/constant.SW_PARENTCLOSING.html

It is possible that this is a widespread issue to other unrelated constants that share the same prefix.
These are just the ones that I was able to find.

Are these errors, or am I misunderstanding something?

riverar:
Fixes #1460

### DIFF
--- a/scripts/ChangesSinceLastRelease.txt
+++ b/scripts/ChangesSinceLastRelease.txt
@@ -198,3 +198,31 @@ Windows.Win32.UI.Shell.IShellWindows.FindWindowSW : swClass...Int32 => ShellWind
 Windows.Win32.UI.Shell.IShellWindows.FindWindowSW : swfwOptions...Int32 => ShellWindowFindWindowOptions
 Windows.Win32.UI.Shell.IShellWindows.Register : swClass...Int32 => ShellWindowTypeConstants
 Windows.Win32.UI.Shell.IShellWindows.RegisterPending : swClass...Int32 => ShellWindowTypeConstants
+# Untangle window scroll/status enums and flags
+Windows.Win32.Media.DirectShow.IVideoWindow.get_WindowState : WindowState...Int32* => SHOW_WINDOW_CMD*
+Windows.Win32.System.Ole.IPropertyPage.Show : nCmdShow...UInt32 => SHOW_WINDOW_CMD
+Windows.Win32.UI.Controls.RichEdit.ITextHost.TxScrollWindowEx : fuScroll...SHOW_WINDOW_CMD => SCROLL_WINDOW_FLAGS
+Windows.Win32.UI.Shell.IShellLinkA.GetShowCmd : piShowCmd...Int32* => SHOW_WINDOW_CMD*
+Windows.Win32.UI.Shell.IShellLinkA.SetShowCmd : iShowCmd...Int32 => SHOW_WINDOW_CMD
+Windows.Win32.UI.Shell.IShellLinkW.GetShowCmd : piShowCmd...Int32* => SHOW_WINDOW_CMD*
+Windows.Win32.UI.Shell.IShellLinkW.SetShowCmd : iShowCmd...Int32 => SHOW_WINDOW_CMD
+Windows.Win32.UI.WindowsAndMessaging.Apis.ScrollWindowEx : flags...SHOW_WINDOW_CMD => SCROLL_WINDOW_FLAGS
+Windows.Win32.UI.WindowsAndMessaging.SCROLL_WINDOW_FLAGS added
+Windows.Win32.UI.WindowsAndMessaging.SCROLL_WINDOW_FLAGS.SW_ERASE added
+Windows.Win32.UI.WindowsAndMessaging.SCROLL_WINDOW_FLAGS.SW_INVALIDATE added
+Windows.Win32.UI.WindowsAndMessaging.SCROLL_WINDOW_FLAGS.SW_SCROLLCHILDREN added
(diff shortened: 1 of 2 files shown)


scripts/ChangesSinceLastRelease.txt
```diff
@@ -188,3 +188,33 @@ Windows.Win32.System.Mmc.View.get_CellContents : CellContents...UInt16** => PBST
 Windows.Win32.System.Mmc.View.get_Memento : Memento...UInt16** => PBSTR
 Windows.Win32.System.Ole.Apis.QueryPathOfRegTypeLib : lpbstrPathName...UInt16** => PBSTR
 Windows.Win32.System.Ole.Apis.VarBstrCat : pbstrResult...UInt16** => PBSTR
+# Untangle window scroll/status enums and flags
+Windows.Win32.Media.DirectShow.IVideoWindow.get_WindowState : WindowState...Int32* => SHOW_WINDOW_CMD*
+Windows.Win32.System.Ole.IPropertyPage.Show : nCmdShow...UInt32 => SHOW_WINDOW_CMD
+Windows.Win32.System.SystemServices.SHOW_WINDOW_STATUS added
```
mikebattista, line 194: We shouldn't be adding types to SystemServices. Either add an explicit namespace to the enum definition or add a usage so the namespace is inherited.

riverar: Good catch!

scripts/ChangesSinceLastRelease.txt
```diff
@@ -188,3 +188,33 @@ Windows.Win32.System.Mmc.View.get_CellContents : CellContents...UInt16** => PBST
...
+Windows.Win32.System.SystemServices.SHOW_WINDOW_STATUS.SW_OTHERZOOM added
+Windows.Win32.System.SystemServices.SHOW_WINDOW_STATUS.SW_PARENTCLOSING added
+Windows.Win32.System.SystemServices.SHOW_WINDOW_STATUS.SW_PARENTOPENING added
+Windows.Win32.UI.Controls.RichEdit.ITextHost.TxScrollWindowEx : fuScroll...SHOW_WINDOW_CMD => SCROLL_WINDOW_FLAGS
+Windows.Win32.UI.Shell.IShellLinkA.GetShowCmd : piShowCmd...Int32* => SHOW_WINDOW_CMD*
+Windows.Win32.UI.Shell.IShellLinkA.SetShowCmd : iShowCmd...Int32 => SHOW_WINDOW_CMD
+Windows.Win32.UI.Shell.IShellLinkW.GetShowCmd : piShowCmd...Int32* => SHOW_WINDOW_CMD*
+Windows.Win32.UI.Shell.IShellLinkW.SetShowCmd : iShowCmd...Int32 => SHOW_WINDOW_CMD
+Windows.Win32.UI.WindowsAndMessaging.Apis.ScrollWindowEx : flags...SHOW_WINDOW_CMD => SCROLL_WINDOW_FLAGS
+Windows.Win32.UI.WindowsAndMessaging.Apis.SW_MAX added
```
mikebattista, line 205: Is it expected that this isn't part of any enum?

riverar: Strange, where that even came from?

riverar: Ah I see it in winuser.h. Missing from the docs! Will fix.

mikebattista: The ConstantScraper picked it up from winuser.h. It used to be autoPopulated into the previous enum, but since it wasn't moved to another, it's now showing up as a loose constant. You can see further below in the diff that SW_MAX was removed from the original enum.

mikebattista (approved):

Issue #1488, opened 2023-03-13
Where is an MIT-licensed `Windows.Win32.winmd` file?


dagood:
* https://github.com/microsoft/win32metadata/issues/766

That issue was partially resolved: https://github.com/microsoft/win32metadata/issues/766#issuecomment-1150295736

> * Yes, the NuGet is licensed weirdly (Windows SDK license). You can get the same files at https://github.com/microsoft/win32metadata/tree/main/scripts/BaselineWinmd to workaround that until it's changed.

That directory no longer exists as of https://github.com/microsoft/win32metadata/pull/1056, and I can't find any winmd files in the repo. [Microsoft.Windows.SDK.Win32Metadata/44.0.17-preview](https://www.nuget.org/packages/Microsoft.Windows.SDK.Win32Metadata/44.0.17-preview) still only contains the Windows SDK license.

---

I'm working on using win32metadata to generate Go's Windows syscalls, but the license is a concern:

* https://github.com/golang/go/issues/43838

I believe Go needs to be able to have a reproducible build without online access, so needs to have the winmd file checked into Go repositories. (Or checking in the whole nuget package, etc.) Copying the source code and building from that doesn't seem reasonable to me, given the .NET dependency.

How can Go get the artifacts in a distributable form?

kennykerr:
You can download the winmd files here: https://www.nuget.org/packages/Microsoft.Windows.SDK.Win32Metadata/ 

The nuget package is just a zip file containing the winmd files. 

If you need an explicit MIT/Apache license, the [windows-rs](https://github.com/microsoft/windows-rs) project provides a copy of the [winmd files here](https://github.com/microsoft/windows-rs/tree/master/crates/libs/metadata/default).

Issue #766, opened 2021-12-18
Clarify license and redistributability of bundled WinSDK headers


mstorsjo:
The repo contains a toplevel file LICENSE, indicating that the files are covered by the MIT license (allowing redistribution etc).

The repo also contains a number of headers from WinSDK (under `generation/WinSDK/RecompiledIdlHeaders`). The headers in WinSDK generally aren't redistributable (the WinSDK license doesn't allow redistribution of anything else than specific redistributables). The files here also contain such a license header that don't seem to imply that it's intended to be freely redistributable under the MIT license, e.g. looking at `um/winnt.h`:

```
/*++ BUILD Version: 0072     Increment this if a change has global effects
    
Copyright (c) Microsoft Corporation. All rights reserved.     
```
(Which explicitly says `All rights reserved.` and nothing about the file being redistributable under any OSS license).

Also as noted in #387, the NuGet package form links to the WinSDK license (which doesn't permit redistribution).

Thus, is it possible to make it clearer that all the included files, including the WinSDK headers, now are redistributable according to the MIT license (which seems to be the intent?), in contrast to when receiving those headers from a regular WinSDK installation?

danny-laa:
FYI, MIT License includes an `All rights reserved.` clause too.

mstorsjo:
> FYI, MIT License includes an `All rights reserved.` clause too.

Hmm, at least my reading of it, at https://opensource.org/licenses/MIT, doesn't contain that phrase in the license itself. But anyway, even if the phrase is used... If it's directly followed by the text granting specific permissions, it's a different thing that when the clause stands on its own with no further grants or even references to other license files.

riverar:
@mikebattista Bubbling this up as it has become quite popular.

mikebattista:
We haven't changed the license of the original SDK headers. The license in this repo is for the metadata and tooling around it.

g2p:
This repo contains foundational metadata for various language bindings. The licence files here are clear enough about the licence for the tools, but the metadata is derived from non redistributable headers, and it is not obvious that you relicense it.

slimsag:
> We haven't changed the license of the original SDK headers.

Is this to be interpreted as "the files under generation/WinSDK/RecompiledIdlHeaders in this repository are not open source / MIT licensed"?

slimsag:
@mikebattista @riverar just curious if you could clarify the above still

ChrisDenton:
[...]

PR #1489, opened 2023-03-13
Clarify licensing

riverar:
See also: https://github.com/microsoft/win32metadata/issues/766#issuecomment-1150295736

Fixes: #1488

@mikebattista Do double-check!


mikebattista (approved):

riverar:
@mikebattista Wonder if can in the future just remove all non-MIT bits and put those in a separate repo to help the cause.

Issue #1529, opened 2023-04-09
Missing constant for arm64 architecture


xverge:
On windows-rs, in the Debug module there are processor architecture constants, [however it is lacking a constant for arm64](https://github.com/microsoft/windows-rs/blob/81b7b8cf3925dc589a5bc8dc5b6a750cf81654b2/crates/libs/windows/src/Windows/Win32/System/Diagnostics/Debug/mod.rs#L7718-L7731), which is [in the win32 documentation.](https://learn.microsoft.com/en-us/windows/win32/api/sysinfoapi/ns-sysinfoapi-system_info#:~:text=PROCESSOR_ARCHITECTURE_ARM64)

riverar:
Isn't missing but rather in a different namespace https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/System/SystemServices/constant.PROCESSOR_ARCHITECTURE_ARM64.html

PR #1530, opened 2023-04-10
Consolidate PROCESSOR_ARCHITECTURE values

riverar:
Fixes: #1529

@mikebattista Heads up, this moves several values from `Windows.Win32.System.SystemServices` to `Windows.Win32.System.Diagnostics.Debug`. Is that what we want? (These values are in `winnt.h` and are not debug specific.)


mikebattista:
I would probably change the order of the uses so the enum moves to System.SystemInformation next to SYSTEM_INFO.

riverar:
@mikebattista Agreed. Will change.

generation/WinSDK/enums.json
```diff
@@ -11900,25 +11900,70 @@
     ]
   },
   {
+    "namespace": "Windows.Win32.System.SystemInformation",
```
mikebattista, line 11903: You shouldn't need to specify this explicitly. Just reorder the "uses" below so that SYSTEM_INFO is the first one and the enum should move to that namespace.

riverar: I'm not sure I like that implicit "first usage defines the namespace" behavior over being explicit. But I guess the thinking here is that this enum should _follow_ around the `SYSTEM_INFO` struct. Will make that edit.

mikebattista (approved):

PR #1538, opened 2023-04-14
Differentiate NativeTypedef structs that exist in the headers vs. only in metadata

### LINKED ISSUE #1533: Ideally, there would be a distinct attribute for these "invented" handle types
Ideally, there would be a distinct attribute for these "invented" handle types, like `MEMORYMAPPEDVIEW_HANDLE`, to distinguish them from native types defs that are actually defined in the Windows SDK. Without this, languages like C++ and Rust that aim to preserve the original definitions more faithfully are now stuck not being able to distinguish between something like `PSID` which is actually defined by the Windows SDK and `MEMORYMAPPEDVIEW_HANDLE` which is not. I thought this was what the `NativeTypeDef` attribute was for but perhaps I misunderstood.

_Originally posted by @kennykerr in https://github.com/microsoft/win32metadata/issues/561#issuecomment-1503504854_

mikebattista:
Fixed #1533.

If we agree to just introduce a different attribute for metadata-only typedefs, we need to close on the name. I started with `NoNativeTypedef`. Another option might be `MetadataTypedef`.

Projections that want to use all the typedefs should just check for both attributes. Projections that want to stay true to the Win32 headers can unwrap `MetadataTypedefs`.

We could also merge the two attributes like `Typedef(Native = true)` and `Typedef(Native = false)`.


riverar:
I like the idea of a separate `MetadataTypedefAttribute` instead. This will allow projection authors to quickly filter out this attribute (if desired). The current implementation requires looking up every single Typedef attribute to read/compare against its Native value.

riverar (changes requested):
I like the idea of a separate `MetadataTypedefAttribute` instead. This will allow projection authors to quickly filter out this attribute (if desired). The current implementation requires looking up every single `TypedefAttribute` to read/compare against its `Native` value.

mikebattista:
Can you clarify? These MetadataTypedefs would still be applied to functions, so projections need to inspect an attribute and then unwrap values either way whether it's a different attribute or a single attribute with a property.

riverar:
The `.custom` declaraction on, say, `HANDLE`, after this PR, may look something like:
```
.custom instance void [Windows.Win32.winmd]Windows.Win32.Foundation.Metadata.TypedefAttribute::.ctor(bool) = (
  01 00 01 00 00
)
```

A projection author wishing to ignore non-native typedefs will have fully parse the attribute, perhaps something like:

```
for each attribute in attributes
  if attribute == TypedefAttribute
    if attribute.valueType == bool && getBoolValue(attribute) != true
      continue;
    // ...
```

The proposed alternative eliminates a bunch of per-attribute work by simply using the type name:

```
for each attribute in attributes
  if attribute == MetadataTypedef
    continue;
  // ...
```

mikebattista:
Ok. I'm fine with that to make the processing more efficient. Your earlier comment suggested maybe a different design altogether for the attribute and how it would be applied.

riverar:
@mikebattista Oh didn't mean to imply that. Was just agreeing with your _"Another option might be MetadataTypedef"_ 👍

Issue #740, opened 2021-11-10
Some Win32 interfaces should be marked with an agile attribute


damyanp:
For example, the various DirectX interfaces (the "nano-COM" ones) should all be marked as agile. This would allow the rust projection to mark them as implementing the Send and Sync traits.

sotteson1:
How can I tell which interfaces should be marked agile? Also, is there an attribute I can use to indicate it's agile, or would I need to make one?

damyanp:
I think we'll need to mark them manually.  Maybe something like "everything in d3d12.h should be agile"?

kennykerr:
Any progress on this? We have more customer interest to resolve this here: https://github.com/microsoft/windows-rs/issues/1519

sotteson1:
> I think we'll need to mark them manually. Maybe something like "everything in d3d12.h should be agile"?

Are we safe making that assumption, that all interfaces in d3d12.h are agile? What about the other d3d headers? Can we make the same assumption for those?

damyanp:
Any d3d* header is definitely agile - these are all [Nano-COM](https://en.wikipedia.org/wiki/Component_Object_Model#Nano-COM_(a.k.a_XPCOM)) interfaces.

sotteson1:
This was accidentally closed. I still need to do this one.

marler8997:
I'm having a hard time finding information on what this `Agile` attribute means exactly.  Is anyone able to help me understand?

riverar:
@marler8997 In this case, it denotes the backing Direct3D 12 objects can be used from any apartment/thread. You're not limited to the apartment/thread that created the object. This information is used by projections, such as Rust, to light up async capabilities.

Does that help at all?

marler8997:
Yeah thanks for the explanation.

PR #1591, opened 2023-06-14
Mark `IDXGI*` interfaces as `[Agile]`

### LINKED ISSUE #1588: Should `IDXGI*` types be marked as `[Agile]`?
#740 and #861 marked the `ID3D*` types `[Agile]` to have `Send` and `Sync` in Rust because all these types are (supposedly) synchronized internally and thus thread-safe. Is the same true for `DXGI` types, and if so can we add the same attribute to them too?

MarijnS95:
Fixes #1588

Just like #740/#861 the `IDXGI*` interface classes should be marked as `Agile` so that `windows-rs` implements `Sync` and `Send` marker `trait`s for these `struct`s in Rust, allowing these types to be fearlessly shared across threads, and accessed from multiple threads at once: their implementations [are thread-safe].

[are thread-safe]: https://github.com/microsoft/win32metadata/issues/1588#issuecomment-1591818653


riverar (changes requested):
Do not merge this yet. `[Agile]` is broken right now.

MarijnS95:
@riverar have you found the issue meanwhile?

riverar:
Currently tracking via https://github.com/microsoft/win32metadata/issues/1588; I haven't had available time yet to sit down and pinpoint the issue but it has something to do with recent Documentation attribute and the fix for specifying multiple attributes in the response files. (The bisection only revealed the commit that unmasked the bug, but it appears it was lurking in wait for us.)

Will try to alter space and time to eek out a few extra minutes tonight to debug further.

riverar:
Blocked on https://github.com/microsoft/win32metadata/pull/1599 will request an expedite

mikebattista:
I fixed the regression with #1600. Rebuilding and merging this now.

PR #1597, opened 2023-06-23
Upgrade Agility SDK to 1.711.3-preview

MarijnS95:
The latest preview includes features like the new D3D12 Work Graphs: https://devblogs.microsoft.com/directx/d3d12-work-graphs-preview/


sources/GeneratorSdk/samples/DiaSdk/Microsoft.Dia.Win32Metadata.nuspec
```diff
@@ -17,7 +17,7 @@
     <icon>images\windows.png</icon>
     <repository type="git" url="https://github.com/microsoft/win32metadata.git" />
     <dependencies>
-      <dependency id="Microsoft.Windows.SDK.Win32Metadata" version="49.0.15-preview-g6e729ece55" />
+      <dependency id="Microsoft.Windows.SDK.Win32Metadata" version="53.0.6-preview-g26ed8da0f6" />
```
MarijnS95, line 20: Oh this keeps automatically trickling in every time... How can we get rid of this changing in the background?

riverar: `.\DoAll.ps1` has `-ExcludePackages -ExcludeSamples` parameters that you may find handy.

MarijnS95: Ah, [`CONTRIBUTING.md` was recently updated](https://github.com/microsoft/win32metadata/commit/18f30abbe5ad742a86115f0a9aed823fe14280be) with that hint, thanks!

riverar (dismissed):
Looks good to me, however we may need to at some point consider having metadata _channels_ to separate production APIs from preview APIs.

mikebattista:
Are the previews purely additive or do they have breaking changes? We haven't ingested any preview releases yet.

riverar:
They say in their outbound comms that the APIs are [not final representations](https://devblogs.microsoft.com/directx/agility-sdk-1-711/#:~:text=note%20that%20this%20is%20a%20preview%20sdk) so sounds like they could break.

mikebattista:
@damyanp what do you think about ingesting previews here?

MarijnS95:
Otherwise I'd appreciate it if we can keep this PR open for visibility (until "channels" are supported) and I will keep updating it with new Agility SDK preview drops and we can ultimately merge it once stabilized.

kennykerr:
> I don't think we should ingest these into the mainline - there's definitely a chance of breaking API changes between preview and final versions of the Agility SDK.

Agreed.

mikebattista:
Are you able to ship bindings for preview releases separately?

MarijnS95:
@mikebattista theoretically Rust projects are able to share [preview releases](https://doc.rust-lang.org/cargo/reference/resolver.html#pre-releases) on crates.io, if there were preview releases of `win32metadata`.

mikebattista:
Is it feasible to ship preview metadata for the Agility SDK separately from win32metadata?

mikebattista:
Shipping the stable APIs here makes sense. I also am not inclined to ship preview APIs here hence my questions about your other options.

MarijnS95:
@mikebattista the other options are indeed creating and publishing separate metadata and a Rust crate, or - what I have now - fork the existing win32metadata and windows-rs crates with these preview bindings (which is less work than respinning the previous aooroach). While that works temporarily, we should keep thinking about longer term upstream solutions and make sure there's visibility towards the current efforts exposing these preview bindings.

mikebattista:
Ok. In terms of visibility of this preview stack you've put together, I would recommend promoting it in the Agility SDK repo and corresponding release announcements.

MarijnS95:
@mikebattista we could ask the writers of the release announcement, e.g. @tex3d :)

mikebattista:
I'll leave it to you. I was just suggesting options given you said you wanted to promote visibility of your preview bindings.

Closing this out as we have no intention of merging this into main.

PR #1599, opened 2023-06-24
Revert documentation attribute additions (62a25c42)

riverar:
Documentation remaps are getting picked up over regex remaps (see code below). This PR reverts the added doc attributes until we fix this and add a few tests.

https://github.com/microsoft/win32metadata/blob/81344326b62096e7c7d62bc9905d4258103eef1d/sources/ClangSharpSourceToWinmd/MetadataSyntaxTreeCleaner.cs#L1198-L1219


riverar:
@mikebattista CI doesn't seem to be kicking in (at least visibly here); could use a hand here so we can unblock other PRs. Thanks!

mikebattista:
I fixed this with #1600.

PR #1679, opened 2023-08-15
Initial support for GDI+

riverar:
Fixes #1287

Issue #1287, opened 2022-09-29
`Gdiplus` missing some functions/enums `GdipCreateSolidFill`, `GdipCloneBrush`, `GpStatus`


elachlan:
https://learn.microsoft.com/en-au/windows/win32/api/gdiplustypes/ne-gdiplustypes-status

GdipCreateSolidFill
https://learn.microsoft.com/en-us/windows/win32/gdiplus/-gdiplus-solidbrush-flat

GdipCloneBrush
https://learn.microsoft.com/en-us/windows/win32/gdiplus/-gdiplus-brush-flat

I can't find the direct documentation on it.

Winforms tracking: https://github.com/dotnet/winforms/issues/7468

chenss3:
These functions are in the header gdiplusflat.h, which are purposefully omitted from the scraper since they're C++ classes. I think at this point, there's not much we can do since these APIs reference C++ types. 


https://github.com/microsoft/win32metadata/blob/384800a2ca454ef603cefd893be9f8a535a12e34/scripts/HeaderFileCoverageExceptions.csv#L177

elachlan:
@sotteson1 is there any way around this for win32metadata?

CC: @JeremyKuhne, we might have to add these manually to winforms in the shorter term if we can't get them included here.

kennykerr:
Note that GDI+ has a C++ wrapper library around a flat C API. The latter should be expressible by win32 metadata. The "C++ classes" are only use as opaque pointers and could be represented as `NativeTypeDef` structs.

elachlan:
@mikebattista any idea on this? Kennykerr's idea above sounds like a plan, but I am unsure how to go about it.

mikebattista:
First step would be to add the header to the right Partition and rebuild to see if the C APIs get picked up.

elachlan:
@mikebattista I see you looked into adding the headers, how did you go?

mikebattista:
It became a non-trivial addition. Someone more familiar with that API would have a better time trying to add the APIs and resolve build errors.

https://github.com/microsoft/win32metadata/compare/mikebattista/gdiplusflat is where I left off.

JeremyKuhne:
Fwiw, TerraFX has this stuff wrapped and might give some guidance: https://github.com/terrafx/terrafx.interop.windows/tree/main/generation/Gdiplus

I'm very interested in getting this in so we can flip System.Drawing over to CsWin32.

mikebattista:
[...]

PR #1683, opened 2023-08-18
Revert "Initial support for GDI+ (#1679)"

riverar:
Reverting 0c63f94aea98688b200a3964365aae60095c83e0 to restore branch status. We'll take another whack at GDI+ shortly.


mikebattista (approved):

elachlan:
Should we re-open #1287?

riverar:
@elachlan Oops, thanks for the heads up. Done. Will get that fixed up in the next few days, tricker than normal due to the flat API not being cleanly separate from the C++ goop.

elachlan:
Thanks for looking at it. It does seem a bit of a hassle trying to wrangle object oriented code to functional. Even if it's just the method calls.

mikebattista:
Yes thanks for looking at it. I got stuck trying to add them so was happy to see you got farther than I did.

PR #1691, opened 2023-08-29
Initial support for UCHAR constants

riverar:
Adds support for scraping negative and positive UCHAR constants.

Also fixes: https://github.com/microsoft/wdkmetadata/issues/48


mikebattista:
Can you confirm that the build was clean?

riverar:
It was clean last night, yep. Not sure why the pipelines stopped running...

mikebattista:
There are security issues with running pipelines on forks and I know there are some policy options in ADO that prevent them from running in that case.

mikebattista (approved):

riverar:
@mikebattista Right, I completely forgot. Will use local branches in the future.

PR #1701, opened 2023-09-09
Remove `LIST_VIEW_INSERT_MARK_FLAGS` enum

elachlan:
```
*** Comparing .winmd to last release
Unknown differences found:
Windows.Win32.UI.Controls.Apis.LVIM_AFTER added
Windows.Win32.UI.Controls.LIST_VIEW_INSERT_MARK_FLAGS removed
Windows.Win32.UI.Controls.LIST_VIEW_INSERT_MARK_FLAGS.LVIM_AFTER removed
Windows.Win32.UI.Controls.LVINSERTMARK.dwFlags...Windows.Win32.UI.Controls.LIST_VIEW_INSERT_MARK_FLAGS => System.UInt32
```


riverar (changes requested):
Where did you get this value from? Not seeing it in public headers.

elachlan:
> The insertion point appears after the item specified if the LVIM_AFTER flag is set; otherwise it appears **before** the specified item.

https://learn.microsoft.com/en-us/windows/win32/api/commctrl/ns-commctrl-lvinsertmark

It doesn't explicitly set it in the headers but the documentation says if its not AFTER then its before.

elachlan:
The other thing to consider is that `LIST_VIEW_INSERT_MARK_FLAGS ` enum before this only had one value.

riverar:
The only thing I can find in headers is:

```cpp
#define LVIM_AFTER      0x00000001
```

While I agree that when the flag is not set a logical _before_ occurs, I don't like the idea of metadata making up constants. (Also consider that values 2, 4, 8, etc. will also leave bit 0 unset.)

Will defer to @mikebattista for the final call though!

elachlan:
If we end up deciding not to add it, then I suggest we remove the enum and use the constants.

This is how we currently define it in Winforms:
https://github.com/dotnet/winforms/blob/478c4e1d28e9adf29938c135aedbb31be47289fd/src/System.Windows.Forms.Primitives/src/Interop/ComCtl32/Interop.LVIM.cs

mikebattista:
> If we end up deciding not to add it, then I suggest we remove the enum and use the constants.
> 
> This is how we currently define it in Winforms: https://github.com/dotnet/winforms/blob/478c4e1d28e9adf29938c135aedbb31be47289fd/src/System.Windows.Forms.Primitives/src/Interop/ComCtl32/Interop.LVIM.cs

I would rather remove the enum than create a new artificial enum member.

elachlan:
PR is updated to remove the enum.

riverar (approved):

PR #1738, opened 2023-11-14
AddSAPI Engine-Level Interfaces

### LINKED ISSUE #1736: SAPI Engine-Level Interfaces missing from Metdata
When looking at the [SAPI 5.4 documentation](https://learn.microsoft.com/en-us/previous-versions/windows/desktop/ee125663(v=vs.85)) on Microsoft learn, there are several interfaces listed that are part of the Win32 metdata, such as ISpObjectWithToken, and ISpVoice. However, it looks like the Engine-Level Interfaces, such as ISpTTSEngine and ISpTTSEngineSite are missing from the metadata. Could the engine level interfaces be added somehow?

leonardder:
Fixes #1736.

I must say that I bumped into one obscure issue though. I had to add `#define _SAPI_VER 0x053` explicitly. The default of sapi.h is `#define _SAPI_VER 0x054`, but that seems to result in a scrape result that is almost half the size of when using version 5.3. I wonder why this is.
Note that sapiddk.h includes sapi.h which includes sapi53.h when `_SAPI_VER` is defined as such.


mikebattista:
Awesome. Thanks!

I did notice a bunch of removals when only including sapiddk.h and using the latest version. I'm not familiar enough with the API to know if that's expected or not.

What you have looks good in terms of adding the missing APIs while preserving what's there already.

mikebattista (approved):

leonardder:
Wow, that's quick, thanks!
May I be so bold as to request a release in the near future? I'm working on a Rust project that relies on the missing interfaces.

PR #1583, opened 2023-05-24
Mark `DXGI_SWAP_CHAIN_DESC1::Flags` as `DXGI_SWAP_CHAIN_FLAG` type

MarijnS95:
This is currently a `UINT` followed by a comment stating the actual enumeration type that contains the possible flag bits.

PR #1757, opened 2023-12-08
 Associate DXGI function parameters and struct fields with the corresponding enum type

MarijnS95:
These are more convenient and descriptive to use than plain `UINT`s.


MarijnS95:
Before merging this, maybe we should do a bigger pass over this namespace and check some more `UINT` APIs, though I haven't come across many obvious ones just yet.

MarijnS95:
This is basically a followup of https://github.com/microsoft/win32metadata/pull/1583 where I should have checked for more types (even ones on the same struct of a different version...). Found a few more, in the `Dxgi` namespace by searching for `*flags: u32` and `*mut u32` in `windows-rs`. Only the following "types" are loose constants instead of associated enums, and cannot be used to better type the function signatures.

https://learn.microsoft.com/en-us/windows/win32/direct3ddxgi/dxgi-present
https://learn.microsoft.com/en-us/windows/win32/api/dxgi/nf-dxgi-idxgisurface-map
https://learn.microsoft.com/en-us/windows/win32/direct3ddxgi/dxgi-enum-modes

---

Curiously in the linked PR we had to annotate these via `[AssociatedEnum("…")]`. Same here for struct fields and `UINT`s passed by value, but for `* UINT`s passed as pointer... Is that expected?

generation/WinSDK/emitter.settings.rsp
```diff
@@ -1748,4 +1747,16 @@ WNetUseConnectionA::return=WIN32_ERROR
...
+GlobalMemoryStatusEx::lpBuffer=[In]
+DXGI_ADAPTER_DESC1::Flags=[AssociatedEnum("DXGI_ADAPTER_FLAG")]
+DXGI_ADAPTER_DESC2::Flags=[AssociatedEnum("DXGI_ADAPTER_FLAG")]
+DXGI_DECODE_SWAP_CHAIN_DESC::Flags=[AssociatedEnum("DXGI_SWAP_CHAIN_FLAG")]
+DXGI_SWAP_CHAIN_DESC::Flags=[AssociatedEnum("DXGI_SWAP_CHAIN_FLAG")]
+DXGI_SWAP_CHAIN_DESC1::Flags=[AssociatedEnum("DXGI_SWAP_CHAIN_FLAG")]
+IDXGIDevice4::OfferResources1::Flags=[AssociatedEnum("DXGI_OFFER_RESOURCE_FLAGS")]
+IDXGIOutput3::CheckOverlaySupport::pFlags=DXGI_OVERLAY_SUPPORT_FLAG
+IDXGIOutput4::CheckOverlayColorSpaceSupport::pFlags=DXGI_OVERLAY_COLOR_SPACE_SUPPORT_FLAG
+IDXGIOutput6::CheckHardwareCompositionSupport::pFlags=DXGI_HARDWARE_COMPOSITION_SUPPORT_FLAGS
```
riverar, line 1759: Don't think these are right. The headers/docs show this as

```
typedef 
enum DXGI_OVERLAY_SUPPORT_FLAG
    {
        DXGI_OVERLAY_SUPPORT_FLAG_DIRECT	= 0x1,
        DXGI_OVERLAY_SUPPORT_FLAG_SCALING	= 0x2
    } 	DXGI_OVERLAY_SUPPORT_FLAG;
```

Which is `int` by default. So it has to stay as-is, however I believe you can associate via the `AssociatedEnum` attribute.

MarijnS95: See my question above:

> Curiously in the linked PR we had to annotate these via `[AssociatedEnum("…")]`. Same here for struct fields and `UINT`s passed by value, but **not** for `* UINT`s passed as pointer... Is that expected?

If not for the `p` prefix it wouldn't be visible that `pFlags` here is actually a pointer to an enum that we're changing the type of... And there's no complaint that it _needs_ to use `AssociatedEnum`.

Besides making it `AssociatedEnum`, should there be similar validation like https://github.com/microsoft/win32metadata/pull/1583#issuecomment-1561407892?

riverar: I think you're asking how to use `AssociatedEnum` with pointers (`UINT*` in this case).

@mikebattista Is that something we support?

MarijnS95: Not really, because I haven't tested it yet. Just asking in the first place if it's expected that the too

Then I assume that it is up to the tooling in e.g. `windows-rs` to understand that a `UINT*` field with `#[AssociatedEnum(XXX)]` annotation really means "a _pointer to_ a value of `XXX`" where `XXX` has the same data layout as `UINT`.

riverar: It's expected because the metadata tooling doesn't have a lot of tests.

I don't think bindgen on the Rust side is expecting AssociatedEnum on pointers either.

MarijnS95: Having `AssociatedEnum` on these (out) pointer types breaks windows-rs completely:

https://github.com/Traverse-Research/windows-rs/commit/ea1c32772c4da263b828f2822d0177acd8dadf22

Out values are no longer returned from functions but are passed _by value_ to the function, where _its value_ is cast to a pointer.

Fwiw:

```
Windows.Win32.Graphics.Dxgi.IDXGIOutput3.CheckOverlaySupport : pFlags : [Out] => [AssociatedEnum(DXGI_OVERLAY_SUPPORT_FLAG),Out]
Windows.Win32.Graphics.Dxgi.IDXGIOutput4.CheckOverlayColorSpaceSupport : pFlags : [Out] => [AssociatedEnum(DXGI_OVERLAY_COLOR_SPACE_SUPPORT_FLAG),Out]
Windows.Win32.Graphics.Dxgi.IDXGIOutput6.CheckHardwareCompositionSupport : pFlags : [Out] => [AssociatedEnum(DXGI_HARDWARE_COMPOSITION_SUPPORT_FLAGS),Out]
Windows.Win32.Graphics.Dxgi.IDXGISwapChain3.CheckColorSpaceSupport : pColorSpaceSupport : [Out] => [AssociatedEnum(DXGI_SWAP_CHAIN_COLOR_SPACE_SUPPORT_FLAG),Out]
```

MarijnS95: @mikebattista @riverar @kennykerr first of all happy new year!

Is there any resolution here? Should we go for the apparently-correct thing which is applying `[AssociatedEnum]` to `UINT *` pointers, and then create an issue over at `windows-rs` to track fixing how it handles this in `bindgen`?

MarijnS95: Reported at https://github.com/microsoft/windows-rs/issues/2771.

riverar: @mikebattista Do we want to support adding `AssociatedEnum` attributes to pointers? This would require projection authors to handle these a bit differently for pointers. (That is, ensure the param type is emitted as T* instead of T.)

@AArnott Do you anticipate any issues with that? Rust side will need a small tweak.

AArnott: Thanks for checking. It seems reasonable to me for CsWin32.

mikebattista: If it's ok with projections then it's no problem to add to the metadata.

MarijnS95: The Rust projections requested to ""resolve"" this in the metadata first. At risk of ending up in an infinite loop, it seems like we can just make a choice here and `windows-rs` will follow along with whatever it ends up with.

riverar: The order of operations is generally: metadata proposes a change, downstream stakeholders provide feedback, metadata makes the change, metadata puts out a release, and then projections adopt or ignore. I don't think it was unreasonable to request, by me, that we wait for something concrete to show up in metadata before doing any work.

Let's move forward with the attribute. You'll still need to correct your emitter.settings.rsp in this PR.

MarijnS95: It isn't an unreasonable request. It is however unreasonable for downstream stakeholders to send someone right back to the metadata, without feedback on the case.

riverar: The feedback is gathered here directly and behind the curtain, not via GitHub issues in the projection repositories. I sympathize though, that wasn't clear, thanks for pointing that out!

MarijnS95: Fair enough, it seemed to me as if we were already with one foot in phase 3: metadata is inclined to make a change. Part of the blame is on me though, the intent of https://github.com/microsoft/windows-rs/issues/2771 was to query downstream projection feedback (as this issue had sat stale for a while - and is a PR thread and not a discussion/issue anyway), though I have not made that abundantly clear.

riverar:
I think these are the remaining items that need conversion?

```
IDXGIOutput3::CheckOverlaySupport::pFlags=DXGI_OVERLAY_SUPPORT_FLAG
IDXGIOutput4::CheckOverlayColorSpaceSupport::pFlags=DXGI_OVERLAY_COLOR_SPACE_SUPPORT_FLAG
IDXGIOutput6::CheckHardwareCompositionSupport::pFlags=DXGI_HARDWARE_COMPOSITION_SUPPORT_FLAGS
```

MarijnS95:
@riverar don't forget about:

```
IDXGISwapChain3::CheckColorSpaceSupport::pColorSpaceSupport=DXGI_SWAP_CHAIN_COLOR_SPACE_SUPPORT_FLAG
```

Those four are the only pointer params in this PR.

riverar:
Right, just looking for you to fix up these last three?

```diff
-IDXGIOutput3::CheckOverlaySupport::pFlags=DXGI_OVERLAY_SUPPORT_FLAG
+IDXGIOutput3::CheckOverlaySupport::pFlags=[AssociatedEnum(DXGI_OVERLAY_SUPPORT_FLAG)]

-IDXGIOutput4::CheckOverlayColorSpaceSupport::pFlags=DXGI_OVERLAY_COLOR_SPACE_SUPPORT_FLAG
+IDXGIOutput4::CheckOverlayColorSpaceSupport::pFlags=[AssociatedEnum(DXGI_OVERLAY_COLOR_SPACE_SUPPORT_FLAG)

-IDXGIOutput6::CheckHardwareCompositionSupport::pFlags=DXGI_HARDWARE_COMPOSITION_SUPPORT_FLAGS
+IDXGIOutput6::CheckHardwareCompositionSupport::pFlags=[AssociatedEnum(DXGI_HARDWARE_COMPOSITION_SUPPORT_FLAGS)


```

MarijnS95:
@riverar so that means there's a definitive _yes_ on annotating pointer types?

riverar:
Yup, lets do it 🚀

MarijnS95:
https://github.com/microsoft/win32metadata/pull/1757#issuecomment-1847959669 if you run into any of these: maybe we can turn the constants into enums?

MarijnS95:
@riverar I'll rebase this when you or @mikebattista have reviewed it and are ready to merge it, to prevent constant force-pushes. Looking forward to the heads-up, thanks!

MarijnS95:
> Only the following "types" are loose constants instead of associated enums, and cannot be used to better type the function signatures.
> 
> https://learn.microsoft.com/en-us/windows/win32/direct3ddxgi/dxgi-present
> https://learn.microsoft.com/en-us/windows/win32/api/dxgi/nf-dxgi-idxgisurface-map
> https://learn.microsoft.com/en-us/windows/win32/direct3ddxgi/dxgi-enum-modes

Since I ran into `enums.json` recently, I figured I should just introduce them.

Again, please help me check that I've caught mostly everything!

generation/WinSDK/enums.json
```diff
@@ -34707,6 +34707,143 @@
...
+    "uses": [
+      {
+        "interface": "IDXGIResource",
+        "method": "SetEvictionPriority",
+        "parameter": "EvictionPriority"
+      },
+      {
+        "interface": "IDXGIResource",
+        "method": "GetEvictionPriority",
+        "parameter": "pEvictionPriority"
```
MarijnS95, line 34811: Looking at windows-rs codegen, this generates correctly 🎉. The diff however changes the ABI:

```
Windows.Win32.Graphics.Dxgi.IDXGIResource.GetEvictionPriority : pEvictionPriority...UInt32* => DXGI_RESOURCE_PRIORITY*
```

That isn't desired. Same for all the other non-pointer type changes introduced by the last commit.

scripts/ChangesSinceLastRelease.txt
```diff
@@ -954,3 +954,87 @@ Windows.Win32.Devices.DeviceAndDriverInstallation.SETUP_DI_BUILD_DRIVER_DRIVER_T
...
+Windows.Win32.Graphics.Dxgi.DXGI_RESOURCE_PRIORITY.DXGI_RESOURCE_PRIORITY_MINIMUM added
+Windows.Win32.Graphics.Dxgi.DXGI_RESOURCE_PRIORITY.DXGI_RESOURCE_PRIORITY_NORMAL added
+Windows.Win32.Graphics.Dxgi.IDXGIDecodeSwapChain.PresentBuffer : Flags...UInt32 => DXGI_PRESENT
+Windows.Win32.Graphics.Dxgi.IDXGIFactory.MakeWindowAssociation : Flags...UInt32 => DXGI_MWA
+Windows.Win32.Graphics.Dxgi.IDXGIFactory3.GetCreationFlags : return...UInt32 => DXGI_CREATE_FACTORY
+Windows.Win32.Graphics.Dxgi.IDXGIResource.GetEvictionPriority : pEvictionPriority...UInt32* => DXGI_RESOURCE_PRIORITY*
+Windows.Win32.Graphics.Dxgi.IDXGIResource.SetEvictionPriority : EvictionPriority...UInt32 => DXGI_RESOURCE_PRIORITY
+Windows.Win32.Graphics.Dxgi.IDXGISurface.Map : MapFlags...UInt32 => DXGI_MAP
+Windows.Win32.Graphics.Dxgi.IDXGISwapChain.Present : Flags...UInt32 => DXGI_PRESENT
+Windows.Win32.Graphics.Dxgi.IDXGISwapChain1.Present1 : PresentFlags...UInt32 => DXGI_PRESENT
```
MarijnS95, line 1037: These should all go through `AssociatedEnum`, right? How does one set that up with `uses` in `enums.json`?

generation/WinSDK/enums.json
```diff
@@ -34707,6 +34707,143 @@
...
+      }
+    ]
+  },
+  {
+    "type": "uint",
+    "name": "DXGI_ENUM_MODES",
+    "flags": true,
+    "autoPopulate": {
+      "filter": "DXGI_ENUM_MODES_",
+      "header": "dxgi.h|dxgi1_2.h"
```
MarijnS95, line 34743: I need some help with this enum. Two constants are defined in `dxgi.h` and two more in `dxgi1_2.h`.

I tried playing around with `addUsesTo`, but as the name implies that is only for `uses` and not for `autoPopulate`'ing extra constants.

scripts/ChangesSinceLastRelease.txt
```diff
@@ -954,3 +954,87 @@ Windows.Win32.Devices.DeviceAndDriverInstallation.SETUP_DI_BUILD_DRIVER_DRIVER_T
...
+Windows.Win32.System.SystemServices.DXGI_SHARED_RESOURCE_RW added
+Windows.Win32.System.SystemServices.DXGI_SHARED_RESOURCE_RW.DXGI_SHARED_RESOURCE_READ added
+Windows.Win32.System.SystemServices.DXGI_SHARED_RESOURCE_RW.DXGI_SHARED_RESOURCE_WRITE added
```
MarijnS95, line 1040: Should I set a `namespace` in `enums.json` for this one? I don't see the constants being used anywhere (in favour of the `_OFFER_` variant) and it's ending up in the `SystemServices` namespace...

generation/WinSDK/enums.json
MarijnS95: @mikebattista can I get some help here?

riverar: The tooling doesn't support multiple headers; you'll have to define them manually for now.

scripts/ChangesSinceLastRelease.txt
riverar: Yep, how does this sound?

```json
"namespace": "Windows.Win32.Graphics.Dxgi",
```

MarijnS95: Marked this as resolved already because that's exactly what I ended up pushing a week ago.

All the other questions remain open though.

generation/WinSDK/enums.json
MarijnS95: It seems very trivial to update the tooling to support a `|` or `;` separator.

generation/WinSDK/enums.json
riverar: If that change isn't desired, you'll have to remove it from `uses` in the json above.

generation/WinSDK/enums.json
```diff
@@ -34723,6 +34723,144 @@
...
+        "interface": "IDXGIResource",
+        "method": "GetEvictionPriority",
+        "parameter": "pEvictionPriority"
+      }
+    ]
+  },
+  {
+    "namespace": "Windows.Win32.Graphics.Dxgi",
+    "type": "uint",
+    "name": "DXGI_SHARED_RESOURCE_RW",
```
riverar, line 34834:
```suggestion
    "name": "DXGI_SHARED_RESOURCE_RIGHTS",
```

"Flags" also acceptable.

generation/WinSDK/enums.json
```diff
@@ -34723,6 +34723,144 @@
...
+      {
+        "interface": "IDXGIOutput1",
+        "method": "GetDisplayModeList1",
+        "parameter": "Flags"
+      }
+    ]
+  },
+  {
+    "type": "uint",
+    "name": "DXGI_MWA",
```
riverar, line 34777:
```suggestion
    "name": "DXGI_MWA_FLAGS",
```

generation/WinSDK/enums.json
```diff
@@ -34723,6 +34723,144 @@
...
+      {
+        "interface": "IDXGIFactory",
+        "method": "MakeWindowAssociation",
+        "parameter": "Flags"
+      }
+    ]
+  },
+  {
+    "type": "uint",
+    "name": "DXGI_MAP",
```
riverar, line 34794:
```suggestion
    "name": "DXGI_MAP_FLAGS",
```

sources/MetadataUtils/ConstantsScraper.cs
```diff
@@ -518,7 +518,7 @@ private void ScrapeConstantsFromTraversedFiles(Dictionary<string, string> traver
                             .Where(
                                 e => e.autoPopulate != null &&
                                 !string.IsNullOrEmpty(e.autoPopulate.filter) &&
-                                e.autoPopulate.header.ToLowerInvariant() == currentHeaderName)
+                                e.autoPopulate.header.ToLowerInvariant().Split(';').Contains(currentHeaderName))
```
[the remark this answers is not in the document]
MarijnS95: It is in the sense that I need this to create an enum for `DXGI_ENUM_MODES`. Do you still want me to split both out to a separate PR?

generation/WinSDK/enums.json
MarijnS95: The type is named RW: https://learn.microsoft.com/en-gb/windows/win32/direct3ddxgi/dxgi-shared-resource-rw

Doesn't that confuse users who might try to `ctrl+F` this? And how about the documentation attribute: https://github.com/Microsoft/win32metadata/blob/09cc32d989ccb945458bb059742435f48aa0879d/generation/WinSDK/documentationMappings.rsp#L13835

generation/WinSDK/enums.json
```diff
@@ -34723,6 +34723,144 @@
...
+      {
+        "interface": "IDXGIDecodeSwapChain",
+        "method": "PresentBuffer",
+        "parameter": "Flags"
+      }
+    ]
+  },
+  {
+    "type": "uint",
+    "name": "DXGI_ENUM_MODES",
```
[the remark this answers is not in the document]
MarijnS95: Yeah, seems really awkward. I'll leave this flag untouched until a decision is made.

generation/WinSDK/enums.json
MarijnS95: @riverar I think what I meant to say is in-line with what was discussed above: this should have an `[AssociatedEnum("DXGI_RESOURCE_PRIORITY")]` attribute?

Is that something I should set up via `emitter.settings.rsp` instead of the `uses` list? What about all the other type changes shown in `ChangesSinceLastRelease`?

The same comment/thread is open at https://github.com/microsoft/win32metadata/pull/1757#discussion_r1455979179.

sources/MetadataUtils/ConstantsScraper.cs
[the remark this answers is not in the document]
riverar: Oh gotcha, you're good.

generation/WinSDK/enums.json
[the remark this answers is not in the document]
mikebattista: The docs call the set DXGI_ENUM_MODES. Stick with the documented name.

generation/WinSDK/enums.json
mikebattista: Set AssociatedEnum in emitter settings if you don't want to remap the type. Plenty of examples there.

sources/MetadataUtils/ConstantsScraper.cs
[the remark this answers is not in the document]
mikebattista: This is great. Thanks for adding the support for multiple headers.

generation/WinSDK/enums.json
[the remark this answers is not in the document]
MarijnS95: @mikebattista how about the other fields that @riverar asked me to add `_FLAGS` to, despite not being documented like that?

sources/MetadataUtils/ConstantsScraper.cs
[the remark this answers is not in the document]
MarijnS95: @mikebattista glad you like this (even if a very simple change)! Any other enum constant we can exercise this support on?

generation/WinSDK/enums.json
mikebattista: As long as the underlying type of the enum matches the type where it's applied, you're not changing the ABI. The build will fail if you change the ABI with an enum.

generation/WinSDK/enums.json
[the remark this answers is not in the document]
mikebattista: If there's a documented name already we should use that.

generation/WinSDK/enums.json
MarijnS95: @mikebattista that seems very strange to me. (It was told that the underlying type of an `enum` is `int`](https://github.com/microsoft/win32metadata/pull/1757#discussion_r1426155306) yet as shown in https://github.com/microsoft/win32metadata/pull/1757#discussion_r1455979179 all these types were changed from `UInt32` to an `int`eger based `enum` (which is signed), yet the build did not fail.

It might be useful to automatically have `"uses"` emit a type change _or_ `[AssociatedEnum]` depending on the original type, if such a thing is even possible?

---

Anyway, I _think_ `[AssociatedEnum]` is then the right way to go for all of these?

generation/WinSDK/enums.json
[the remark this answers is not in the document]
MarijnS95: I've reverted the `_FLAGS` suggestion from @riverar for two enums that already have upstream documentation:

- `DXGI_PRESENT`
- `DXGI_SHARED_RESOURCE_RW`

I've left the others like this, and added a documentation URL to a function that describes their values (unsure if this is desired?):

- `DXGI_CREATE_FACTORY_FLAGS`
- `DXGI_MAP_FLAGS`
- `DXGI_MWA_FLAGS`
- `DXGI_RESOURCE_PRIORITY` (no rename was suggested here)

generation/WinSDK/enums.json
MarijnS95: Scratch that, the reply above is completely wrong (talking about `enum` types being `int`, when this PR concerns itself with `#define`s scraped _into_ enums).

What I see in ILSpy for these `enum`s is that they have a `: uint` type, and the `#define`s are all defined with `UL` constant suffix, further clarifying that their type is `UINT`. Accordingly, these `enum`s are the same type as the original `UINT` fields/parameters and can be safely replaced without changing ABI, hence the build succeeding and hence no `AssociatedEnum` being needed.

generation/WinSDK/documentationMappings.rsp
```diff
@@ -13781,6 +13781,7 @@ DXGI_ADAPTER_FLAG3=[Documentation("https://learn.microsoft.com/windows/win32/api
 DXGI_ALPHA_MODE=[Documentation("https://learn.microsoft.com/windows/win32/api/dxgi1_2/ne-dxgi1_2-dxgi_alpha_mode")]
 DXGI_COLOR_SPACE_TYPE=[Documentation("https://learn.microsoft.com/windows/win32/api/dxgicommon/ne-dxgicommon-dxgi_color_space_type")]
 DXGI_COMPUTE_PREEMPTION_GRANULARITY=[Documentation("https://learn.microsoft.com/windows/win32/api/dxgi1_2/ne-dxgi1_2-dxgi_compute_preemption_granularity")]
+DXGI_CREATE_FACTORY_FLAGS=[Documentation("https://learn.microsoft.com/windows/win32/api/dxgi/nf-dxgi-createdxgifactory")]
```
MarijnS95, line 13784: @mikebattista any thought about linking the relevant function page if there's no type documentation available at all? Or should I revert all these additions to `documentationMappings.rsp`?

riverar: I'd delete them. Bindings may generate incorrect summaries/comments.

generation/WinSDK/enums.json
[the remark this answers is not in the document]
MarijnS95: @riverar any comments here? This looks to be the last issue to resolve (thus far), and I'd love to get this change in before the next `win32metadata` release to reap the benefits together with the Agility SDK 613 regen 😄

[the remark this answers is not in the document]
riverar: Checking those last four now, one moment.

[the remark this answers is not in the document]
riverar: tl;dr: LGTM. If any of the "no existing enumeration" assertions below are invalid, please let me know!

Re: `DXGI_CREATE_FACTORY_*`, these are documented as _flags_ to be passed into CreateDXGIFactoryN functions. Currently its only value is: `#define DXGI_CREATE_FACTORY_DEBUG 0x1`. There is no existing enumeration that captures these constants, therefore I propose: `DXGI_CREATE_FACTORY_FLAGS`.

Re: `DXGI_MAP_*`, these are documented as _flags_ to be passed into IDXGISurface::Map and possibly others. It has values such as DXGI_MAP_READ, [...]WRITE, and [...]DISCARD. There is no existing enumeration that captures these constants, therefore I propose: `DXGI_MAP_FLAGS`. This would also align with how we handle similar flags, such as `REG_SAM_FLAGS`.

Re: `DXGI_MWA_*`, these are documented as _flags_ to be passed into IDXGIFactory::MakeWindowAssociation. It has values such as DXGI_MWA_NO_ALT_ENTER. There is no existing enumeration that captures these constants, therefore I propose: `DXGI_MWA_FLAGS`.

Re: `DXGI_RESOURCE_PRIORITY`, these are documented as _mutually exclusive constants_ to be passed into IDXGIResource::SetEvictionPriority. It has values such as DXGI_RESOURCE_PRIORITY_LOW. There is no existing enumeration that captures these constants, therefore I propose keeping the original name: `DXGI_RESOURCE_PRIORITY`.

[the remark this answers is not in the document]
MarijnS95: Thanks for the detailed clarification, that makes total sense!

Let's mark this thread as resolved.

riverar (approved):

MarijnS95:
Thanks for the review @riverar! @mikebattista let me know when this is ready to be merged, so that I can solve the conflicts with `ChangesSinceLastRelease.txt`.

mikebattista (approved):

mikebattista:
Merged. Thanks!

generation/WinSDK/Partitions/Dxgi/settings.rsp
```diff
@@ -13,3 +13,11 @@ _D3DCOLORVALUE
 <IncludeRoot>/um/dxprogrammablecapture.h
 --namespace
 Windows.Win32.Graphics.Dxgi
+--with-attribute
+DXGI_MULTIPLANE_OVERLAY_YCbCr_FLAGS=Flags
+DXGI_OFFER_RESOURCE_FLAGS=Flags
+DXGI_OUTDUPL_FLAG=Flags
```
MarijnS95, line 19: This flag has one `DXGI_OUTDUPL_COMPOSITED_UI_CAPTURE_ONLY = 1` constant, but I couldn't find in what API it is supposed to be used. Only the constructor for output duplication has as `Flags` field that "should always be zero":


https://learn.microsoft.com/en-us/windows/win32/api/dxgi1_5/nf-dxgi1_5-idxgioutput5-duplicateoutput1

riverar: Looks like a question for the DX repo. Suspect this is was meant to stay internal. But a real world example from OS components verifies your claim: `pObject->DuplicateOutput1(device, DXGI_OUTDUPL_COMPOSITED_UI_CAPTURE_ONLY, ...)`

MarijnS95: Sounds good, in that case the documentation should be updated too.

What's the desired way to assign the enum, emitter.settings.rsp or listing it in enums.json?

riverar: Think we prefer `enums.json` these days!

MarijnS95: @riverar that might mean we could move more of the `--with-attribute` here - with explicit reassignments in `emitter.settings.rsp` - and consolidate them into `enums.json`?

Looks like there's also `--enumMakeFlags` in `emitter.settings.rsp` doing a similar thing 😅

MarijnS95: > But a real world example from OS components verifies your claim: `pObject->DuplicateOutput1(device, DXGI_OUTDUPL_COMPOSITED_UI_CAPTURE_ONLY, ...)`

@riverar where'd you find this example?

riverar: `bcastdvruserservice.dll!Windows::Media::Capture::Internal::DDACapture::AcquireOutputDuplication` (26217.5000), reverse engineered 😅

MarijnS95: Posted the question at https://github.com/microsoft/DirectX-Headers/issues/128, let's see.

PR #1769, opened 2023-12-18
Add missing NotifyIcon notification constant

### LINKED ISSUE #1765: `NIN_KEYSELECT` missing
The constant `NIN_KEYSELECT` is missing (for `Shell_NotifyIconW()`).

riverar:
Fixes: #1765


mikebattista:
What's blocking it from being int?

riverar:
@mikebattista I didn't fully explore that rabbit hole, but I think we'll need to change WM_USER and all the NIN_ msgs to int and then fix all the collateral damage. 😅

riverar:
Interesting, made the change locally again and doesn't seem as bad. Will update this PR shortly.

riverar:
Looks like AzDO has given up on builds. @mikebattista

mikebattista:
All the other WM_ constants are uint. Are there side effects of making WM_USER int but not the others? Feels like this should cause some usability problems somewhere.

riverar:
Yeah, I think there's a larger "naked constants should default to int not uint" discussion to be had here.

mikebattista:
> Looks like AzDO has given up on builds. @mikebattista

Thanks. I believe it should be fixed now.

software-artificer:
Is anything stopping this from being merged? Just ran into missing `NIN_KEYSELECT` in `windows-rs`.

PR #1824, opened 2024-01-17
Only load netstandard.dll from the hardcoded path if it wasn't provided by the user

dpaoliello:
## Issue Details:
`ClangSharpSourceToWinmd` currently uses a hard-coded path to find the `netstandard.dll` reference assembly, which is problematic if the user has a project-local version (e.g., using the [`NETStandard.Library.Ref` NuGet package](https://www.nuget.org/packages/NETStandard.Library.Ref)) or otherwise has it in a non-standard location.

## Fix Details:
Only use the hard-coded path if the compilation doesn't already have netstandard 2.1 already in the compilation: this allows users to specify a path using the normal `--ref` command line argument but falls back to the legacy behavior for customers who depend on that.

### DIFF
--- a/sources/ClangSharpSourceToWinmd/ClangSharpSourceCompilation.cs
+++ b/sources/ClangSharpSourceToWinmd/ClangSharpSourceCompilation.cs
@@ -143,22 +143,7 @@ namespace ClangSharpSourceToWinmd
             string objDir = Path.Combine(topObjDir, arch);
             Directory.CreateDirectory(objDir);
 
-            var netstandardPath = FindNetstandardDllPath();
-            if (!File.Exists(netstandardPath))
-            {
-                throw new FileNotFoundException("Failed to find the netstandard DLL.");
-            }
-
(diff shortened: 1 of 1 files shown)


riverar:
Hm, would be great to avoid adding some random variable that folks will depend on and we'll have to support forever. Maybe we can come up with a better scheme to find this library in the first place?

dpaoliello:
@riverar I've changed my approach for this: instead of using an env var, what if we only inject netstandard.dll if the user didn't already add it via `--ref`?

riverar:
@dpaoliello Sounds promising; think your last commit got mangled there and can no longer see a diff.

dpaoliello:
> Sounds promising; think your last commit got mangled there and can no longer see a diff.

Yeah, sorry, something messed up the line endings - should be good now.

riverar (approved):
Nice!

mikebattista (approved):

Issue #1566, opened 2023-05-08
`StgConvertPropertyToVariant` is missing


kennykerr:
Tested the latest version of Win32/Wdk and this function has disappeared, blocking ingestion. Any ideas?

mikebattista:
It was removed as part of https://github.com/microsoft/win32metadata/issues/1547 since it depended on a C++ class PMemoryAllocator that was being emitted as an empty struct. C++ APIs are typically excluded.

In this case, should a class like this be treated like a `typedef struct` opaque handle? Or is excluding the right thing?

kennykerr:
Ugh, the `IMemoryAllocator` definition was never published publicly so it's probably better to avoid it.

riverar:
@mikebattista @kennykerr

Am wondering if we should revisit this. `IMemoryAllocator` is documented (https://learn.microsoft.com/en-us/windows/win32/stg/imemoryallocator) so calling this API should be possible.

And the function parameter is documented as
> A pointer to a class that implements the [IMemoryAllocator](https://learn.microsoft.com/en-us/windows/desktop/Stg/imemoryallocator) abstract class.

I recognize the headers still refer to abstract class `PMemoryAllocator` but it seems to align with the interface as expected.

```cpp
class PMemoryAllocator
{
public:
    virtual void *Allocate(ULONG cbSize) = 0;
    virtual void Free(void *pv) = 0;
};
...
```

kennykerr:
Yes, even if the parameter is just declared as `void*` it would at least make the API available for those who need it.

riverar:
Propose we:
* add a manual definition for `IMemoryAllocator`, based on [docs](https://learn.microsoft.com/windows/win32/stg/imemoryallocator) and validated against `PMemoryAllocator`

   ```cpp
    IMemoryAllocator
    {
    public:
      virtual void* Allocate(ULONG cbSize) = 0;
      virtual void Free(void* pv) = 0;
    }
   ```
* restore `StgConvertPropertyToVariant` and point to interface above
* consider filing a SDK bug to get the interface defined in headers to align with docs

PR #1832, opened 2024-01-20
Restore StgConvertPropertyToVariant, adjust param type

riverar:
Fixes: #1566


mikebattista (changes requested):
Thanks. Just update the baseline then we can merge.

mikebattista (approved):

PR #1850, opened 2024-02-21
fix: IoRing return types

localcc:
This commit fixes the IoRing function return types. `PopIoRingCompletion` function has multiple success values which should be reflected in the metadata.


mikebattista (approved):

mikebattista:
Thanks!

PR #1877, opened 2024-03-29
Move Azure build pipelines to OneBranch

getrou:
As part of compliance with updated security requirements, production pipelines must run in a 1ES environment. To do this, pipelines are converting to use OneBranch pipeline templates.

This changes how artifacts are published and the build environment is slightly different, resulting in some small changes to the build files.

As a backup plan, the existing azure pipeline scripts are left in place. These will be removed in the future.


AzurePipelinesTemplates/win32metadata-onebranch.yml
```diff
@@ -0,0 +1,200 @@
...
+      parameters:
+        RepoDirectory: ${{ parameters.RepoDirectory }}
+
+    - task: UseDotNet@2
+      displayName: ⚙ Install .NET SDK
+      inputs:
+        packageType: sdk
+        useGlobalJson: true
+
+    # TODO - this no longer sets an Azure Devops pipeline build number, OneBranch has a different way of doing that
```
mikebattista, line 46: Is this TODO still pending?

getrou: It is, this Azure Devops pipeline name doesn't get properly set right now

mikebattista: The current release pipeline uses the pipeline build number so I don't think we can merge this until this is resolved.

riverar:
Does it make sense to migrate the pipeline to GitHub Actions and call out to a backend for signing?

mikebattista:
No. It looks like we still have access to the existing pipelines, though, so I plan to push out a release this week.

AArnott:
@georou What is your plan for a release pipeline? I don't see one introduced by this PR.

FWIW, for all the 1ES PT pipeline migrations I've done, I've just reused the existing YAML files for the 1ES PT variety, which kept me from having to create new pipelines in AzDO. It also avoided having YAML files for deprecated pipelines that could confuse folks who want to look at the new pipeline files.

riverar:
Still unclear to me why this isn't just a normal GHA workflow.

AArnott:
@riverar put simply, Microsoft has compliance requirements for shipping software that to date is only implemented on Azure Pipelines, AFAIK.

riverar:
Are we shipping _software_ in this repository though? It's a bunch of non-executable metadata. 😂 (I understand Microsoft is sensitive to this right now though!)

mikebattista:
Yes we ship binary tools in addition to metadata and NuGet packages that are signed by Microsoft. GHA doesn't support our workflows.

getrou:
> @georou What is your plan for a release pipeline? I don't see one introduced by this PR.
> 
> FWIW, for all the 1ES PT pipeline migrations I've done, I've just reused the existing YAML files for the 1ES PT variety, which kept me from having to create new pipelines in AzDO. It also avoided having YAML files for deprecated pipelines that could confuse folks who want to look at the new pipeline files.

All these pipelines now run in the github-private/microsoft ADO project, which has its own (very slim) repo associated with it, and uses this github repository as a pipeline resource. The release pipeline YAML has moved there, but we could move it here if we want, I think.

PR #1886, opened 2024-04-12
Add references for all assemblies passed in via `--ref`

dpaoliello:
While trying to generate my own bindings, `ClangSharpSourceToWinmd` was reporting the following error:

```
error CS0648: '' is a type not supported by the language
```

I tracked it down to a method in the current compilation using a struct from a referenced WinMD (let's call this "Middle.winmd") which had a field from another WinMD (let's call it "Base.winmd"). Looking at the winmd files in ildasm, Middle.winmd was missing a reference to Base.winmd, so there was no way for the current compilation to know where referenced types came from, despite the current compilation referencing both Middle.winmd and Base.winmd.

The fix for this is to add an assembly reference to all assemblies passed in via `--ref`

### DIFF
--- a/sources/ClangSharpSourceToWinmd/ClangSharpSourceWinmdGenerator.cs
+++ b/sources/ClangSharpSourceToWinmd/ClangSharpSourceWinmdGenerator.cs
@@ -144,31 +144,17 @@ namespace ClangSharpSourceToWinmd
 
             void InitReferences()
             {
-                Version systemVersion = new Version(2, 1, 0, 0);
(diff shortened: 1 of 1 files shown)


riverar (approved):

kennykerr:
@mikebattista?

mikebattista (approved):

mikebattista:
Thanks for the contribution.

Our build and release pipelines are being [refactored](https://github.com/microsoft/win32metadata/pull/1877) which is slowing things down. Once that's completed, we can resume updates.

Issue #1891, opened 2024-04-30
`RAIIFree` implies `InvalidHandleValue`


kennykerr:
Taking another stab at supporting `RAIIFree` in Rust, but the following structs have the `RAIIFree` attribute but lack at least one `InvalidHandleValue` attribute. Is that intentional? Are we meant to assume that zero is thus the invalid value?

* HIORING
* PTP_CLEANUP_GROUP
* PTP_POOL
* SOCKET

kennykerr:
Conversely, there are many structs that have at least one `InvalidHandleValue` attribute but lack a corresponding `RAIIFree` attribute. Is that intentional? 

- HCONN
- HCOMDB
- ADS_SEARCH_HANDLE
- HIFTIMESTAMPCHANGE
- HAMSICONTEXT
- HAMSISESSION
- JET_LS
- JET_OSSNAPID
- OLE_HANDLE
- HRSRC
- PSID
- HACCESSOR
- BCRYPT_ALG_HANDLE
- BCRYPT_HANDLE
- PRJ_DIR_ENTRY_BUFFER_HANDLE
- PRJ_NAMESPACE_VIRTUALIZATION_CONTEXT
- HCS_CALLBACK
- JET_API_PTR
- JET_HANDLE
- JET_INSTANCE
- JET_SESID
- JET_TABLEID
- HDC
- HGDIOBJ
- HMONITOR
- HCERTSTORE
- HCERTSTOREPROV
- HCRYPTPROV_LEGACY
- HCRYPTPROV_OR_NCRYPT_KEY_HANDLE
- NCRYPT_HANDLE
- NCRYPT_HASH_HANDLE
- NCRYPT_SECRET_HANDLE
- HIMC
- HIMCC
- HSAVEDUILANGUAGES
- CO_DEVICE_CATALOG_COOKIE
- CO_MTA_USAGE_COOKIE
- LPM_HANDLE
- RHANDLE
- HRASCONN
- HTASK
- HPSS
- HPSSWALK
- HACMDRIVER
- HACMDRIVERID
- HACMOBJ
- HACMSTREAM
- HMIDI
- HMIDIIN
- HMIDIOUT
- HMIDISTRM
- HMIXER
- HMIXEROBJ
- HWAVE
- HWAVEIN
- HWAVEOUT
- HCONV
- HCONVLIST
- HDDEDATA
- HSZ
- HTTP_PUSH_WAIT_HANDLE
- WEB_SOCKET_HANDLE
[...]

PR #1895, opened 2024-04-30
Add additional RAII metadata

riverar:
Closes: #1891

Adds additional RAII supporting user-mode functions that take one required parameter, and optionally a second optional nullable parameter. All others are documented as unsupported (at this time) in the mentioned issue.


riverar:
Not likely to compile, as a lot of the functions are not in the same type namespace. This will have to be fixed.

```
EMITWINMD : error : Failed to create source for auto types: PSID, HKL, HDIAGNOSTIC_DATA_QUERY_SESSION, HDIAGNOSTIC_REPORT, HDIAGNOSTIC_EVENT_TAG_D
ESCRIPTION, HDIAGNOSTIC_EVENT_PRODUCER_DESCRIPTION, HDIAGNOSTIC_EVENT_CATEGORY_DESCRIPTION, HDIAGNOSTIC_RECORD, FH_SERVICE_PIPE_HANDLE, HIMC, HPTP
ROVIDER, ROPARAMIIDHANDLE, JET_INSTANCE, JET_SESID not in the same namespace as CloseApi. CloseApi relationships must exist in the same namespace.
```

riverar:
Moved the aforementioned types closer to their creation/destruction functions. This is a breaking change, heads up @mikebattista @kennykerr @AArnott

generation/WinSDK/autoTypes.json
```diff
@@ -80,8 +80,9 @@
     "NativeTypedef": true
   },
   {
-    "Namespace": "Windows.Win32.Foundation",
+    "Namespace": "Windows.Win32.Security",
```
mikebattista, line 83: PSID was probably in Foundation to avoid circular namespace dependencies. Did we regress anything here?

riverar: Unknown. Tests are green and I ran this through Rust bindings with no ill effects. (We'll need to update WDK metadata.) I don't think metadata guarantees non-circular namespace dependencies, does it?

mikebattista: @kennykerr are you concerned about this change at all? We have a test for cyclical namespaces, so if that passed then it might be ok, but would be good to sanity check on your end.

kennykerr: Yep all good on my end.

riverar: An alternative, if this becomes an issue, is to move the functions into Foundation. But that seems wrong given there's nothing foundational about SIDs.

kennykerr: Right, it's a security API. The cycle thing never amounted to much. I wouldn't worry about it.

mikebattista (dismissed):

PR #1915, opened 2024-05-29
Add Cryptographic Provider Development Kit APIs, types

### LINKED ISSUE #1907: Need Rust bindings for bcrypt_provider.h and ncrypt_provider.h
My team is writing a KSP Provider in Rust. Currently we need to use the following headers in our implementation of a KSP provider:

bcrypt.h
ncrypt.h
wincrypt.h
bcrypt_provider.h
ncrypt_provider.h
 
Based on our analysis, the first 3 headers are already covered by the windows-rs crate, but it seems that the last 2 headers, for the providers specifically, are not included. It would be great to get those eventually included to allow us to avoid our own bindgen completely for writing a KSP provider. Please let me know if you need any additional info or if there's anything else I can do to help. (Issue posted here after discussing with @jder )

riverar:
Fixes #1907.

Add Cryptographic Provider Development Kit (CPDK) Version 8.0 APIs, types to metadata.


mikebattista (changes requested):
You need to add the headers to the AdditionalHeaders folder as well, otherwise they'll be lost the next time the SDK headers are recompiled or we upgrade the SDK.

The relevant step in the build process is below.

https://github.com/microsoft/win32metadata/blob/8d678c59bf1c69089e52567086094ae0e01c2c1c/scripts/RecompileIdlFilesForScraping.ps1#L44-L45.

riverar:
@mikebattista Thanks, done!

generation/WinSDK/Partitions/Security.Cryptography/settings.rsp
```diff
@@ -15,5 +16,53 @@ _LM_OWF_PASSWORD
 <IncludeRoot>/um/i_cryptasn1tls.h
 <IncludeRoot>/um/infocard.h
 <IncludeRoot>/um/mssign.h
+<IncludeRoot>/cpdk/bcrypt_provider.h
```
mikebattista, line 19: The recompile script I linked recursively moves everything under AdditionalHeaders to the `um` directory in the `<IncludeRoot>` folder, so I believe these `cpdk` paths need to be changed to `um` to resolve properly.

riverar: Ah right, so a subsequent build after IDL recompile will fail. Will change that now.

mikebattista (approved):

mikebattista:
Thanks!

PR #1919, opened 2024-06-03
When finding a type def from a type ref, search all assemblies

dpaoliello:
We were seeing a crash in `ClangSharpSourceToWinmd` when we had an interface inheriting another interface from a different assembly:
```
3001>Unhandled exception: System.InvalidOperationException: Sequence contains no matching element
3001>   at System.Linq.ThrowHelper.ThrowNoMatchException()
3001>   at System.Linq.Enumerable.First[TSource](IEnumerable`1 source, Func`2 predicate)
3001>   at MetadataUtils.WinmdUtils.InternalGetImplementedMethodCount(TypeDefinition typeDef)
3001>   at MetadataUtils.WinmdUtils.GetTypes()+MoveNext()
3001>   at ClangSharpSourceToWinmd.ClangSharpSourceWinmdGenerator.<PopulateMetadataBuilder>g__GatherInterfaceTypes|60_1()
3001>   at ClangSharpSourceToWinmd.ClangSharpSourceWinmdGenerator.PopulateMetadataBuilder()
3001>   at ClangSharpSourceToWinmd.ClangSharpSourceWinmdGenerator.GenerateWindmdForCompilation(ClangSharpSourceCompilation compilation, Dictionary`2 typeImports, HashSet`1 reducePointerLevels, HashSet`1 forceGuidConsts, Version version, String outputFileName)
3001>   at ClangSharpSourceToWinmd.Program.Run(InvocationContext context)
```

`ClangSharpSourceToWinmd` was only searching for matching named types within the current assembly when it encountered a `TypeReference`, however `TypeReference` is specifically used when referencing a type from a foreign assembly.

This fix for this is to provide the complete set of metadata readers to `GetTypes` so that it can search through all of them to find the type definition for a given `TypeReference`

PR #1921, opened 2024-06-06
Fix test failures

dpaoliello:
PR #1919 introduced a test failure since it was passing an empty list for `allWinmds`, fix is to pass a list containing the current WinMD.


mikebattista (approved):

mikebattista:
Thanks.

Issue #1941, opened 2024-06-28
IPropertyStoreCapabilities has no way to return its value


asklar:
## Actual behavior
IPropertyStoreCapabilities::IsPropertyWritable returns S_OK or S_FALSE
However it is projected as returning void

## Expected behavior

ideally return true/false

AArnott:
In the meantime, @asklar, you can use the NativeMethods.json file to add this method to the preserveSig methods so that you get the HRESULT back.

riverar:
Fixed for next release.  As Andrew mentioned, an easy short term workaround is to disable CsWin32 signature transformation for this method, as such:

```json
{
    "$schema": "https://aka.ms/CsWin32.schema.json",
    "comInterop": {
        "preserveSigMethods": [
            "IPropertyStoreCapabilities.IsPropertyWritable"
        ]
    }
}
```

PR #1942, opened 2024-07-05
Add attribute on IPropertyStoreCapabilities::IsPropertyWritable

riverar:
Fixes: #1941

Adds `CanReturnMultipleSuccessValues` to `IPropertyStoreCapabilities::IsPropertyWritable` method. This method can return either `S_OK` or `S_FALSE`. Also removes the optional Attribute suffix on a few entries as cleanup.


AArnott (dismissed):
Thanks!

mikebattista (changes requested):
You don't need the "-Attribute" suffix here and we don't use it elsewhere in this file. Please remove it for consistency.

riverar:
Oh I cut/pasted it from earlier in the file (GetMessage). Will fix.

mikebattista (approved):
Thanks.

PR #1970, opened 2024-08-27
Annotate missing `ComOutPtr` in `CompositionSwapchain` and `DirectCom…

MarijnS95:
…position`

`dcomp.h` and `Presentation.h` are missing some `_COM_Outptr_` annotations, turning these functions in `windows-rs` to take a meaningless `*mut *mut c_void`.  With this annotation the functions return a generic `T: Interface` _and_ automatically pass its `&T::IID` to the `riid` parameter resulting in a much more ergonomic API.

Also annotate some return-only functions with `RetVal` to make it easier to retrieve these objects.


mikebattista (approved):

mikebattista:
Thanks!

generation/WinSDK/emitter.settings.rsp
```diff
@@ -986,6 +986,13 @@ IWICImagingFactory::CreateDecoderFromFilename::pguidVendor=[Optional]
 DCompositionCreateDevice::dcompositionDevice=[ComOutPtr]
 DCompositionCreateDevice2::dcompositionDevice=[ComOutPtr]
 DCompositionCreateDevice3::dcompositionDevice=[ComOutPtr]
+DCompositionGetStatistics::frameStats=[RetVal]
+DCompositionGetTargetStatistics::targetStats=[RetVal]
```
riverar, line 990: The `RetVal` attribute is only used to indicate that a particular parameter should be treated as the return value of a method. Its use on a Win32 API doesn't really make sense, especially this one where there are multiple outputs, and may even cause downstream codegen issues.

```cpp
STDAPI DCompositionGetStatistics(
    _In_ COMPOSITION_FRAME_ID frameId,
    _Out_ COMPOSITION_FRAME_STATS* frameStats,
    _In_ UINT targetIdCount,
    _Out_writes_opt_(targetCount) COMPOSITION_TARGET_ID* targetIds,
    _Out_opt_ UINT* actualTargetIdCount);
```

riverar: (@MarijnS95 Tagging you just in case notifications aren't firing, due to the PR status.)

kennykerr: Yep, I don't believe this will work in Rust and should be reverted.

mikebattista:
Thanks for the feedback. @MarijnS95 could you please file a separate PR to remove the RetVal changes?

generation/WinSDK/emitter.settings.rsp
MarijnS95: @riverar notifications never get stuck like that. I did receive this message but end up not doing sub-second replies :)

Correct this `RetVal` did not change the `windows-rs` API for `DCompositionGetStatistics()`, but it did (for the better!) for `DCompositionGetTargetStatistics()`. Since the former is the only one of these with auxiliary output data (the thing you seem to have a problem with), that should be the only annotation to revert.

riverar: That attribute is for methods only, not functions, so both will need to be reverted.

MarijnS95: #1980.

MarijnS95: > That attribute is for methods only, not functions, so both will need to be reverted.

Then maybe fix `windows-rs` first to not handle the attribute on functions? It definitely has a positive impact:

```diff
diff --git a/crates/libs/windows/src/Windows/Win32/Graphics/DirectComposition/mod.rs b/crates/libs/windows/src/Windows/Win32/Graphics/DirectComposition/mod.rs
index 9e0f070ac..6bf150323 100644
--- a/crates/libs/windows/src/Windows/Win32/Graphics/DirectComposition/mod.rs
+++ b/crates/libs/windows/src/Windows/Win32/Graphics/DirectComposition/mod.rs
@@ -76,9 +76,10 @@ pub unsafe fn DCompositionGetStatistics(frameid: u64, framestats: *mut COMPOSITI
     DCompositionGetStatistics(frameid, framestats, targetidcount, core::mem::transmute(targetids.unwrap_or(std::ptr::null_mut())), core::mem::transmute(actualtargetidcount.unwrap_or(std::ptr::null_mut()))).ok()
 }
 #[inline]
-pub unsafe fn DCompositionGetTargetStatistics(frameid: u64, targetid: *const COMPOSITION_TARGET_ID, targetstats: *mut COMPOSITION_TARGET_STATS) -> windows_core::Result<()> {
+pub unsafe fn DCompositionGetTargetStatistics(frameid: u64, targetid: *const COMPOSITION_TARGET_ID) -> windows_core::Result<COMPOSITION_TARGET_STATS> {
     windows_targets::link!("dcomp.dll" "system" fn DCompositionGetTargetStatistics(frameid : u64, targetid : *const COMPOSITION_TARGET_ID, targetstats : *mut COMPOSITION_TARGET_STATS) -> windows_core::HRESULT);
-    DCompositionGetTargetStatistics(frameid, targetid, targetstats).ok()
+    let mut result__ = core::mem::zeroed();
+    DCompositionGetTargetStatistics(frameid, targetid, &mut result__).map(|| result__)
 }
 #[inline]
 pub unsafe fn DCompositionWaitForCompositorClock(handles: Option<&[super::super::Foundation::HANDLE]>, timeoutinms: u32) -> u32 {
```

Issue #1453, opened 2023-02-06
Bug: ID3D12GraphicsCommandList::ClearRenderTargetView pRects should be marked optional


chyyran:
https://learn.microsoft.com/en-us/windows/win32/api/d3d12/nf-d3d12-id3d12graphicscommandlist-clearrendertargetview

`pRects` is not indicated as optional

```cpp
void ClearRenderTargetView(
  [in] D3D12_CPU_DESCRIPTOR_HANDLE RenderTargetView,
  [in] const FLOAT [4]             ColorRGBA,
  [in] UINT                        NumRects,
  [in] const D3D12_RECT            *pRects
);
```

However in the documentation

> [in] pRects
> 
> Type: const D3D12_RECT*
>
> An array of D3D12_RECT structures for the rectangles in the resource view to clear. If NULL, ClearRenderTargetView clears the entire resource view.

Because it isn't marked optional, it is impossible to pass in `None` in windows-rs so the `NULL` behaviour is not obtainable.

MarijnS95:
1.5 years later: it looks like you forgot about [`ID3D12GraphicsCommandList::ClearDepthStencilView()`](https://learn.microsoft.com/en-us/windows/win32/api/d3d12/nf-d3d12-id3d12graphicscommandlist-cleardepthstencilview), which has the same semantics.

https://github.com/microsoft/win32metadata/pull/1971

PR #1971, opened 2024-08-27
Mark `ID3D12GraphicsCommandList::ClearDepthStencilView` `pRects` as `Optional`

MarijnS95:
In #1453 `ClearRenderTargetView()`'s `pRect` slice was requested to be marked as optional, but its `ClearDepthStencilView()` sibling was completely glossed over.  Mark it `Optional` as well as a `NULL` pointer is expected to clear the entire resource view.  A non-`NULL` pointer with a length of `0` might not clear anything at all.


MarijnS95:
@kennykerr is this something we can get in a `windows-rs` release?

mikebattista (approved):

mikebattista:
Thanks!

Issue #1961, opened 2024-08-09
Few little errors


adeyblue:
1 - windows.win32.networkmanagement.rras
PMPRADMINCONNECTIONHANGUPNOTIFICATION3
Last parameter is missing a pointer, see https://learn.microsoft.com/en-us/windows/win32/api/mprapi/nf-mprapi-mpradminconnectionhangupnotification3
The SDK is also missing the pointer in mprapi.h,. That struct is relatively big so I doubt it's meant to be passed by value, especically as it's a pointer in all the other function pointers that use RAS_CONNECTION_3 and has members tagged as OUT in mprapi.

2 - Links for all the AVI- prefixed types in Windows.Win32.Media.DirectShow are 404
For instance:
AVIStreamHeader
https://learn.microsoft.com/windows/win32/api/avifmt/ns-avifmt-avistreamheader
This is now here - https://learn.microsoft.com/en-us/previous-versions/windows/desktop/api/avifmt/ns-avifmt-avistreamheader
AVISUPERINDEX
https://learn.microsoft.com/windows/win32/api/aviriff/ns-aviriff-avisuperindex
This is now here - https://learn.microsoft.com/en-us/previous-versions/windows/desktop/api/aviriff/ns-aviriff-avisuperindex
AVISTREAMHEADER
https://learn.microsoft.com/windows/win32/api/aviriff/ns-aviriff-avistreamheader
Now here - https://learn.microsoft.com/en-us/previous-versions/windows/desktop/api/aviriff/ns-aviriff-avistreamheader

Basically they've all moved into previous-versions purgatory

3- These shouldn't have the StructSizeField attribute, cbSize here is the size of the data pointed to, not the size of the struct
windows.win32.com.structuredstorage - BLOB
windows.win32.storage.structuredstorage - BSTRBLOB

4 - Help links for windows.win32.graphics.gdiplus BitMap and Image are pointing to completely wrong things
Bitmap is here
https://learn.microsoft.com/en-us/windows/win32/api/gdiplusheaders/nl-gdiplusheaders-bitmap
instead of
https://learn.microsoft.com/windows/win32/tablet/bitmap-element

Image is here
https://learn.microsoft.com/en-us/windows/win32/api/gdiplusheaders/nl-gdiplusheaders-image
instead of
https://learn.microsoft.com/windows/win32/windowsribbon/windowsribbon-element-image

5 - windows.win32.networking.winhttp
[...]

PR #1978, opened 2024-08-29
Remove StructSizeField from BLOB, BSTRBLOB

riverar:
Related: #1961

Removed `StructSizeField` from `BLOB` and `BSTRBLOB` as these were mistakenly flagged due to the `cbSize` fields. The logic that adds this attribute has also been moved up to allow proper remapping/removal of this attribute.


mikebattista (approved):

mikebattista:
Thanks!

PR #1981, opened 2024-08-29
dxgi: Add DXGI_OUTDUPL_FLAG enum annotation to DuplicateOutput1()

MarijnS95:
Draft separation from https://github.com/microsoft/win32metadata/pull/1911#issuecomment-2319149565.

On the side, alphabetically sort some flags and use `--enumMakeFlags` in `emitter.settings.rsp` instead of `--with-attribute XXX=Flags` in `Namespace/settings.rsp`.


generation/WinSDK/Partitions/Dxgi/settings.rsp
```diff
@@ -13,11 +13,3 @@ _D3DCOLORVALUE
 <IncludeRoot>/um/dxprogrammablecapture.h
 --namespace
 Windows.Win32.Graphics.Dxgi
---with-attribute
```
MarijnS95, line 16: Inversely, could we also have all the `=AssociatedEnum` in here? It seems much nicer to me to have configuration in the namespace-specific `settings.rsp` than in one humongous global file, if possible?

mikebattista: No. The scraper and the emitter have different roles and supporting toolchains.

mikebattista (dismissed):

generation/WinSDK/Partitions/Dxgi/settings.rsp
mikebattista: https://github.com/microsoft/win32metadata/blob/main/docs/architecture.md

MarijnS95: @mikebattista unfortunately that doesn't explain directly which flags are available, nor what precedence I should stick to. Searching further in the repo shows that `--enumMakeFlags` is a feature of `ClangSharpSourceToWinmd` so it makes sense that that only works via emitter settings.

Then the question remains: do we prefer to set the `Flags` attribute on the _emitter_ via `--enumMakeFlags`, or do we prefer to set it _earlier_ via `--with-attribute` on the _scraper_ as done before this PR?

MarijnS95: @mikebattista are you able to help with the above question, so that we can further this PR?

riverar: Resolving this thread as it seems like a tangent and unrelated to the PR.

riverar (dismissed):

riverar (dismissed):

PR #1989, opened 2024-08-30
DXC: Annotate new struct fields and enums

MarijnS95:
> [!CRITICAL]
> Unfinished, untested draft for https://github.com/microsoft/win32metadata/pull/474#issuecomment-2322000133.

### DIFF
--- a/generation/WinSDK/enums.json
+++ b/generation/WinSDK/enums.json
@@ -28759,20 +28759,6 @@
       "filter": "DXC_CP_",
       "header": "dxcapi.h"
     },
-    "members": [
-      {
-        "name": "DXC_CP_ACP",
-        "value": "0"
-      },
-      {
-        "name": "DXC_CP_UTF16",
-        "value": "1200"
-      },
-      {
-        "name": "DXC_CP_UTF8",
-        "value": "65001"
-      }
-    ],
     "uses": [
       {
         "interface": "IDxcBlobEncoding",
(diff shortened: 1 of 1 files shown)


generation/WinSDK/enums.json
```diff
@@ -28818,6 +28804,132 @@
...
+        "method": "ValidateWithDebug",
+        "parameter": "Flags"
+      }
+    ]
+  },
+  {
+    "name": "DxcVersionInfoFlags",
+    "flags": true,
+    "autoPopulate": {
+      "filter": "DxcVersionInfoFlags_",
```
MarijnS95, line 28838: Unfortunately both these `static const UINT32`s don't appear to be parsed by `enums.json`?

mikebattista: Seems more like a ClangSharp limitation than enums.json.

MarijnS95: Yeah, either way it's something that I expected to show up, but it didn't. Should I manually declare `"members"`, while leaving `"autoPopulate"` in place in hopes of getting a compilation conflict when ClangSharp can provide these to us, or is it a fix you can do elsewhere in the tooling?

As this PR shows we used to have manual fields for `DXC_CP_` too (now unnecessary), while also relying on `autoPopulate` without conflicts being generated?

mikebattista: Yes you can manually declare the missing constants together with autoPopulate. The sets will be merged.

MarijnS95: Cool. Fixed it by removing it, because they can be autodetected.

I don't remember the status of this PR: I think I got stuck on the `static const UINT32`s not being parsed, only `#define`s? Is there any plan/solution for that?

MarijnS95: > Seems more like a ClangSharp limitation than enums.json.

Am I right to assume that `ClangSharp` emits whatever it finds _after preprocessing_? I'd assume that all the preprocessor directives are flattened out, which is why the `ConstantsScraper` specifically looks for `#define` in source files:

https://github.com/microsoft/win32metadata/blob/2e6b61977d4b7f5f852637bb7c2d121d1d687a96/sources/MetadataUtils/ConstantsScraper.cs#L32-L34

This is the only code that looks at `autoPopulate`. Assuming that "actual C constants" are extracted from ClangSharp and written elsewhere (could it be [`ConstantWriter`](https://github.com/microsoft/win32metadata/blob/main/sources/MetadataUtils/ConstantWriter.cs)?), is this where we should insert extra logic to turn `static const` definitions into grouped `enum`s as well?

riverar: If I remember correctly, I was previously working on another issue and was also surprised that all constants emitted by Clangsharp were dropped on the floor and only manually defined/ConstantsScraper constants were actually used in the final product. It's not an ideal situation.

So I'd manually define those values for now in the json (sorry) until the larger issue of constants is resolved.

MarijnS95: Thanks @riverar, I've defined them manually now!

PR #1992, opened 2024-09-03
ScrapeHeaders: Trim filenames when splitting for `;`

### LINKED ISSUE #1838: Trim whitespace when parsing `<TraverseFiles>` contents?
For https://github.com/Tsukisoft/direct-storage-rs/pull/5

We had a list of the form:

```xml
            <TraverseFiles>
                $(PkgMicrosoft_Direct3D_DirectStorage)\native\include\dstorage.h;
                $(PkgMicrosoft_Direct3D_DirectStorage)\native\include\dstorageerr.h
            </TraverseFiles>
```

But no constants from `dstorageerr.h` were parsed. Swapping the filenames makes it so that `#define`s from `dstorage.h` are lost, but defines from `dstorageerr.h` are parsed.

I believe the raw string it's parsing here is: `dstorage.h;\ndstorageerr.h\n`, splitting this on `;` causes the last filename to never match. Perhaps a `string item = items[i].Trim();` is missing here to make multiline filename strings work more naturally?

https://github.com/microsoft/win32metadata/blob/3d2ba1e501e46f078e4b4670a52ed68890870f8a/sources/GeneratorSdk/MetadataTasks/ScrapeHeaders.cs#L476-L488

After all, when merging this into one line:

```xml
            <TraverseFiles>$(PkgMicrosoft_Direct3D_DirectStorage)\native\include\dstorage.h;$(PkgMicrosoft_Direct3D_DirectStorage)\native\include\dstorageerr.h</TraverseFiles>
```

Both files were finally parsed correctly.

MarijnS95:
Fixes #1838


MarijnS95:
TODO: Need to check how many more things need this.

riverar (approved):

MarijnS95:
@mikebattista no, unfortunately I have not set up the repro case yet nor validated all the other `.Split()` calls that **seem like they _could_** benefit from the same change but am not sure about. Not to speak of the many remaining untouched `.Split()` calls. Guess I'm in a bit over my head.

MarijnS95:
In addition, I don't actually remember (if I ever did this at all) how to turn `<Project Sdk="Microsoft.Windows.WinmdGenerator/0.59.13-preview">` into something that imports or replaces the "Sdk" with my own (local checkout of this) PR?

riverar:
I just hit this again in another project, very annoying. Let's get this merged in and we can address any issues if they come up.

PR #2073, opened 2025-03-10
Generate `dxdiag.h` bindings in `DirectX.Diagnostics` namespace

MarijnS95:


generation/WinSDK/emitter.settings.rsp
```diff
@@ -2020,4 +2020,10 @@ CreateTransaction::IsolationLevel=[In][Optional]
 CreateTransaction::IsolationFlags=[In][Optional]
 CreateTransaction::Timeout=[In][Optional]
 CreateStreamOnHGlobal::hGlobal=[Optional]
-D3D11CreateDevice::Software=[Optional]
\ No newline at end of file
+D3D11CreateDevice::Software=[Optional]
+IDxDiagProvider::Initialize::pParams=[Const]
+IDxDiagContainer::GetNumberOfChildContainers::pdwCount=[Out]
+IDxDiagContainer::EnumChildContainerNames::pwszContainer=[Out][NativeArrayInfo(CountFieldName = "cchContainer")]
+IDxDiagContainer::EnumPropNames::pwszPropName=[Out][NativeArrayInfo(CountFieldName = "cchPropName")]
```
MarijnS95, line 2027: Noting to self that this needs to use `CountParamIndex` so that this special piece of code in `windows-rs` translates the function argument to `&mut [u16]` (unfortunately loosing the `PWSTR` type) to pass the "capacity" of the _output_ string into `cch*` without users having to do so manually:

https://github.com/microsoft/windows-rs/blob/4523c9d2c12b2227f369b3ed86ceb0ba9db409d2/crates/libs/bindgen/src/types/cpp_method.rs#L632-L640

Since there are a few more functions with `CountFieldName` attributes, maybe those should be renamed or windows-rs needs to lookup by name as well?

(Bonus: `PWSTR` can't be used as function argument because it defines the length based on `strlen()` that searches for a NUL-terminator, which isn't defined for an arbitrary "output only" buffer. The transparent `PWSTR` pointer wrapper "newtype" doesn't behave like an "owned" Rust `Vec` which carries a length and capacity field)

tests/Windows.Win32.Tests/assets/suggestedRemappingsAllowList.rsp
```diff
@@ -258,4 +258,5 @@ Gdiplus::BrushType=GpBrushType
 Gdiplus::MatrixOrder=GpMatrixOrder
 Gdiplus::FlushIntention=GpFlushIntention
 Gdiplus::PathData=GpPathData
-_KNONVOLATILE_CONTEXT_POINTERS_ARM64=KNONVOLATILE_CONTEXT_POINTERS_ARM64
\ No newline at end of file
+_KNONVOLATILE_CONTEXT_POINTERS_ARM64=KNONVOLATILE_CONTEXT_POINTERS_ARM64
+_DXDIAG_INIT_PARAMS=DXDIAG_INIT_PARAMS
```
riverar, line 262: This should be remapped in your partition, not ignored (allowed). Add:

```
--remap
_DXDIAG_INIT_PARAMS=DXDIAG_INIT_PARAMS
```

to `settings.rsp` in your partition.

riverar (changes requested):

riverar (dismissed):

tests/Windows.Win32.Tests/assets/suggestedRemappingsAllowList.rsp
riverar: Yeah, I fall into this trap myself too. The warning makes it very easy to do the wrong thing.

Your previous remap is fine as-is, but to answer your question: yes probably. I'm not sure what the status of metadata is at the moment, so I would hold off on adding another PR.

PR #2088, opened 2025-05-31
Expose NtCreateNamedPipeFile

andfoy:
This PR adds `NtCreateNamedPipeFile` (https://learn.microsoft.com/en-us/windows/win32/devnotes/nt-create-named-pipe-file) to the list of available binding signatures, this function is used to provide support for asynchronous access for pseudo-terminals via ConHost: https://github.com/microsoft/terminal/blob/2d64a3a4ab0d0d1c5b9085a022f5319460ef68d7/src/types/utils.cpp#L851

### DIFF
--- a/generation/WinSDK/RecompiledIdlHeaders/um/winternl.h
+++ b/generation/WinSDK/RecompiledIdlHeaders/um/winternl.h
@@ -487,6 +487,29 @@ NtOpenFile (
 
 //
 // use the Win32 API instead
+//     CreateNamedPipeA
+//
+__kernel_entry NTSTATUS
+NTAPI
+NtCreateNamedPipeFile(
+  OUT PHANDLE FileHandle,
+  IN ULONG DesiredAccess,
+  IN POBJECT_ATTRIBUTES ObjectAttributes,
+  OUT PIO_STATUS_BLOCK IoStatusBlock,
+  IN ULONG ShareAccess,
+  IN ULONG CreateDisposition,
+  IN ULONG CreateOptions,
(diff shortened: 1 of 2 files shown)


generation/WinSDK/RecompiledIdlHeaders/um/winternl.h
```diff
@@ -485,6 +485,29 @@ NtOpenFile (
     IN ULONG OpenOptions
     );
 
+//
+// use the Win32 API instead
+//     CreateNamedPipeA
+//
```
riverar, line 491: This comment won't bubble up anywhere useful so you can delete it.

riverar (approved):

riverar (approved):

jonwis (approved):

generation/WinSDK/RecompiledIdlHeaders/um/winternl.h
```diff
@@ -485,6 +482,29 @@ NtOpenFile (
...
+__kernel_entry NTSTATUS
+NTAPI
+NtCreateNamedPipeFile(
```
jevansaks: Won't this file get clobbered the next time we run UpdateSDK with whatever is really in the SDK?

is this function definition available in a Windows SDK header that we could use / include instead?

riverar: Docs say it lives in `ntioapi.h` which isn't public. Agree, we should minimally move it to a new header (e.g., `\generation\WinSDK\AdditionalHeaders\ntioapi.h`)

PR #2120, opened 2025-09-15
Add missing WLDP and SPOOLSS APIs

jevansaks:
These APIs are not declared as extern "C" but they are exported as un-mangled name from the dll so for p/invoke they should be treated as un-mangled. Looks like other headers have solved this by wrapping the header includes with extern "C" so doing the same here.

Also SDK 22000 is no longer available to be installed from VS installer so bump to 26100 for the dev environment requirements.


mikebattista (changes requested):
Can you build and update the baseline? I'm assuming this adds APIs to the metadata.

Issue #486, opened 2021-05-18
FARPROC shouldn't be arch-specific


AArnott:
The `FARPROC` delegate is defined uniquely based on architecture:

```diff
-[SupportedArchitecture(Windows.Win32.Interop.Architecture.X64 | Windows.Win32.Interop.Architecture.Arm64)]
+[SupportedArchitecture(Windows.Win32.Interop.Architecture.X86)]
-public delegate IntPtr FARPROC();
+public delegate int FARPROC();
```

But using `IntPtr` for both would be equivalent, and allow AnyCPU targeted assemblies to consume the API.

AArnott:
Is there a way to catch all similar APIs to fix in bulk?

sotteson1:
Looks like there are three that are defined at the same place using this pattern:

````C
#ifdef _WIN64
typedef INT_PTR (FAR WINAPI *FARPROC)();
typedef INT_PTR (NEAR WINAPI *NEARPROC)();
typedef INT_PTR (WINAPI *PROC)();
#else
typedef int (FAR WINAPI *FARPROC)();
typedef int (NEAR WINAPI *NEARPROC)();
typedef int (WINAPI *PROC)();
````

PR #2124, opened 2025-09-26
Update crossarch list based on new component partitions

jevansaks:
IP6_ADDRESS used to be per-arch but was dropped when it moved to the Dns partition (which was set as ExcludeFromCrossArch). Remove Dns partition from ExcludeFromCrossArch so we can once again have per-arch IP6_ADDRESS.


scripts/ChangesSinceLastRelease.txt
```diff
@@ -0,0 +1,16 @@
+# Adding *PROC, IP6_ADDRESS, EapHostPeerGetDataToUnplumbCredentials back to arch-specific
+Windows.Win32.Foundation.FARPROC removed
+Windows.Win32.Foundation.FARPROC(X64, Arm64) added
+Windows.Win32.Foundation.FARPROC(X86) added
+Windows.Win32.Foundation.NEARPROC removed
+Windows.Win32.Foundation.NEARPROC(X64, Arm64) added
+Windows.Win32.Foundation.NEARPROC(X86) added
+Windows.Win32.Foundation.PROC removed
+Windows.Win32.Foundation.PROC(X64, Arm64) added
+Windows.Win32.Foundation.PROC(X86) added
```
riverar, line 10: I don't think this is right. We represent `FARPROC` for example as an `IntPtr` adorned with `[UnmanagedFunctionPointer(CallingConvention.Winapi)]` to cover all supported archs already.

This will probably break you @AArnott, just a heads up. (https://github.com/microsoft/win32metadata/issues/486)

scripts/ChangesSinceLastRelease.txt
```diff
@@ -0,0 +1,16 @@
...
+Windows.Win32.NetworkManagement.Dns.IP6_ADDRESS removed
+Windows.Win32.NetworkManagement.Dns.IP6_ADDRESS(X64, Arm64) added
+Windows.Win32.NetworkManagement.Dns.IP6_ADDRESS(X86) added
+Windows.Win32.Security.ExtensibleAuthenticationProtocol.Apis.EapHostPeerGetDataToUnplumbCredentials removed
+Windows.Win32.Security.ExtensibleAuthenticationProtocol.Apis.EapHostPeerGetDataToUnplumbCredentials(X64, Arm64) added
```
riverar, line 15: These don't look right either. I think it's tripping on `phCredentialImpersonationToken`, of which we currently represent with an `IntPtr`.

riverar (changes requested):

scripts/ChangesSinceLastRelease.txt
jevansaks: Ok, again there is something wrong with the winmdgenerator tool then.

riverar (commented):
The headers show:

```
typedef union
{
#ifdef _WIN64
    QWORD       IP6Qword[2];
#endif
    DWORD       IP6Dword[4];
    WORD        IP6Word[8];
    BYTE        IP6Byte[16];
#ifdef  IN6_ADDR
    IN6_ADDR    In6;
#endif
}
IP6_ADDRESS, *PIP6_ADDRESS;
```

Metadata currently represents the IP6_ADDRESS as:

```
[StructLayout(LayoutKind.Explicit)]
public struct IP6_ADDRESS
{
	[FieldOffset(0)]
	public uint[] IP6Dword;

	[FieldOffset(0)]
	public ushort[] IP6Word;

	[FieldOffset(0)]
	public byte[] IP6Byte;
}
```

This structure exists to facilitate type punning of IPv6 addresses in C++. The union allows the same 16-byte IPv6 address to be accessed as different integer types without copying data.

I'm not sure this PR is worth the cost of downstream churn of multiple architecture-specific types. For example, folks may be using this in AnyCPU apps and will suddenly need to do a bunch of work for a QWORD accessor they likely don't care about.

I think a better path forward for improving `IP6_ADDRESS` could be:
* Keep the type architecture-neutral
* Add the missing QWORD member
* Add the missing IN6_ADDR member

This approach provides the missing functionality without breaking existing code or forcing architecture-specific handling.

AArnott:
I'm concerned that we're thinking only about IP6_ADDRESS here. Isn't it the whole header file? Unless anyone has audited the file carefully, we should assume there may be other `#if` sections that are arch-specific and thus should preserve that in the metadata. 
If we know that this one type is the only one that is arch-specific, then I'd consent to leaving architecture turned off, since as you say, it's more convenient to be AnyCPU compatible.

But as @jevansaks says, converting it to be arch-specific will hardly be a horrible blow to people, since it has been arch-specific for a very long time.

riverar (dismissed):

riverar:
Agreed! When I realized it was a regression, it definitely changed the calculus here.

jevansaks:
Thanks @riverar! @mikebattista, can you approve & merge?

PR #2125, opened 2025-09-27
Fix return value of ShellExecute

### LINKED ISSUE #2123: ShellExecute{W,A} should be annotated with `[return: DoNotReleaseAttribute]`
In the metadata, `HINSTANCE` is attributed such that CsWin32 will produce a `SafeHandle` that calls `FreeLibrary`. 

```cs
[RAIIFree("FreeLibrary")]
[AlsoUsableFor("HMODULE")]
[InvalidHandleValue(0L)]
[NativeTypedef]
public struct HINSTANCE
{
	public unsafe void* Value;
}
```

The docs say that `ShellExecute` doesn't really return a valid HINSTANCE value: [ShellExecuteW function (shellapi.h) - Win32 apps | Microsoft Learn](https://learn.microsoft.com/en-us/windows/win32/api/shellapi/nf-shellapi-shellexecutew#return-value)

> If the function succeeds, it returns a value greater than 32. If the function fails, it returns an error value that indicates the cause of the failure. The return value is cast as an HINSTANCE for backward compatibility with 16-bit Windows applications. It is not a true HINSTANCE, however. It can be cast only to an INT_PTR and compared to either 32 or the following error codes below.

Therefore, we shouldn't call FreeLibrary on it, I suppose.

In that case, the metadata is missing a `[return: DoNotReleaseAttribute]` on the ShellExecute method, which calls out the special case, and then CsWin32 will not emit the SafeHandle as a return type.

jevansaks:
Fixes #2123


AArnott (approved):

sources/GeneratorSdk/samples/DiaSdk/Microsoft.Dia.Win32Metadata.nuspec
```diff
@@ -17,7 +17,7 @@
     <icon>images\windows.png</icon>
     <repository type="git" url="https://github.com/microsoft/win32metadata.git" />
     <dependencies>
-      <dependency id="Microsoft.Windows.SDK.Win32Metadata" version="54.0.28-preview" />
+      <dependency id="Microsoft.Windows.SDK.Win32Metadata" version="65.0.1-preview-gfdd0306498" />
```
riverar, line 20: This isn't related to ShellExecute, revert?

riverar (changes requested):

sources/GeneratorSdk/samples/DiaSdk/Microsoft.Dia.Win32Metadata.nuspec
jevansaks: The build keeps changing it. I don't know why. Maybe this file should be .gitignored?

riverar: It likes you! (Will check it out.)

jevansaks: DoAll.ps1 script runs this which scribbles on that file. Should the file be ignored or should this scribbling happen into an obj dir? https://github.com/microsoft/win32metadata/blob/main/scripts/UpdateSampleDependencies.ps1

@mikebattista looks like this came from https://github.com/microsoft/win32metadata/commit/67b8c7e945a884cd0b4ad6adbf9c74b309610a71#diff-48ea1c3e7e64d76be8daf15d44d92b2ea237f91849b86f46c06fce0f803fb7f7 . Any ideas what the solution should be so I don't keep getting this file edited in my local repo? :)

riverar: Oh, for inner-loop dev, I would add `-ExcludeSamples -ExcludePackages`.

mikebattista (approved):

PR #2133, opened 2025-10-15
Add [Const] annotation for [In] string parameters in Printer APIs

jevansaks:
Based on documentation and looking at some of the implementations, it is clear that these _In_ PWSTR params are actually constant. To make these APIs nicer to use in cswin32, I'm proposing to add [Const] to these params so they project as `string` paremeters.

I used Copilot to help grab the list of APIs and parameters that needed this annotation so it's possible I missed some, but I spot-checked and it seems right. :)


riverar (approved):
I can't speak to the implementation side but looks reasonable to me.

PR #2139, opened 2025-10-22
Update UpdateSdk scripts so that running the script on the same version is idempotent

jevansaks:
I was finding there were a lot of manual steps in the process of updating to a new SDK so I've gone through and fixed a bunch of issues. Now if you run `.\scripts\UpdateSDK.ps1 10.0.26100.2161` (the version we're on), there will be no deltas that you have to manually revert.

I did a bunch of things:
1. Moved manual libMappings to their own file, libMappingsManual.rsp.
2. Running UpdateSDK was populating a bunch of libMappings.rsp in a different order than before. I'm not sure how we got to the previous order, maybe there's some non-stable sort happening? But in any case running the tool repeatedly gets the same output, so this seems correct.
3. Moved all manual additions (things that have been removed from the SDK or that are documented as GetProcAddress only APIs) to libMappingsManual.rsp.
4. RecompileIdlforScraping.ps1 was not handling the case where MIDL fails with error output but returns "success" error code. We need to look at the output.
5. This uncovered that a few headers like oledb.idl were not compiling due to them needing to compile with /amd64 instead of /win32. I switched all headers over to compile MIDL /amd64.
6. Win32Metadata wants to include headers which have since been deleted from the SDK. I added an exclusion list to the RecompileIdlFilesForScraping script which restores these headers.
7. WinHVEmulation headers and friends were lost in a recent SDK update. I'm going to try to fix the Windows SDK side but for now, leave those untouched as well.
8. Update DiaSDK.proj to not rewrite the nuspec file, instead pass the win32metadata version variable through as a nuget parameter.
9. Many SAL attributes were being manually added back to oledb.h. This is because ConvertMidlAttributesToSalAnnotations.ps1 is not handling more complex parenthesized expressions like "size_is((ULONG)count)". I've updated the script to handle these better. (The SAL handling is still pretty messed up, but at least this keeps things consistent with before).

For context, there is an oddity of the win32metadata repo currently where it's a bizarre amalgam of new SDK and old SDKs, so things that have been removed from the Windows SDK are still in Win32Metadata. I think we should do something about that, but I've preserved that behavior for now.

### DIFF
--- a/scripts/ChangesSinceLastRelease.txt
+++ b/scripts/ChangesSinceLastRelease.txt
@@ -1,0 +1,4 @@
+
+# Fix headers that have gotten out of sync with SDK
+Windows.Win32.System.Search.IRowsetKeys.ListKeys : prgColumns : [NativeArrayInfo(CountParamIndex=0),Out] => [In,NativeArrayInfo(CountParamIndex=0),Out]
+Windows.Win32.System.Search.IColumnsInfo.GetColumnInfo : pcColumns : [Out] => [Optional,Out]
--- a/generation/WinSDK/emitter.settings.rsp
+++ b/generation/WinSDK/emitter.settings.rsp
@@ -2122,7 +2122,6 @@ FindFirstPrinterChangeNotification::return=FINDPRINTERCHANGENOTIFICATION_HANDLE
 FindNextPrinterChangeNotification::hChange=FINDPRINTERCHANGENOTIFICATION_HANDLE
 FindClosePrinterChangeNotification::hChange=FINDPRINTERCHANGENOTIFICATION_HANDLE
 # Changes to make oledb.h functions consistent with old annotations
-IColumnsInfo::GetColumnInfo::pcColumns=[-Optional]
 IRowsetKeys::ListKeys::prgColumns=[-In][NativeArrayInfo(CountParamIndex=0)]
-IScopedOperations::Copy::pAuthenticate=[-Optional]
-IScopedOperations::Move::pAuthenticate=[-Optional]
+IScopedOperations::Copy::pAuthenticate=[Optional]
+IScopedOperations::Move::pAuthenticate=[Optional]


scripts/ChangesSinceLastRelease.txt
```diff
@@ -0,0 +1,5 @@
+# Getting OLE headers in sync with the SDK again after fixing MIDL issue
+Windows.Win32.System.Search.IColumnsInfo.GetColumnInfo : pcColumns : [Out] => [Optional,Out]
+Windows.Win32.System.Search.IRowsetKeys.ListKeys : prgColumns : [NativeArrayInfo(CountParamIndex=0),Out] => [In,Out]
+Windows.Win32.System.Search.IScopedOperations.Copy : pAuthenticate : [In,Optional] => [In]
+Windows.Win32.System.Search.IScopedOperations.Move : pAuthenticate : [In,Optional] => [In]
```
riverar: These two should remain optional
https://learn.microsoft.com/en-us/previous-versions/windows/desktop/ms711633(v=vs.85)
https://learn.microsoft.com/en-us/previous-versions/windows/desktop/ms720893(v=vs.85)

scripts/RecompileIdlFilesForScraping.ps1
```diff
@@ -82,4 +96,91 @@ $idlFilesToRecompile | ForEach-Object -ThrottleLimit ([System.Math]::Max([System
...
+    "um\qcc\platform.h",
+    "um\qcc\windows\mapping.h",
+    "um\qcc\windows\platform_types.h",
+    "um\windows.devices.alljoyn.interop.h",
+    "um\windows.devices.alljoyn.interop.idl",
+    "winrt\windows.devices.alljoyn.h",
+    "winrt\windows.devices.alljoyn.idl"
+)
+
+Write-Host "Restoring deprecated headers that are no longer in the SDK..."
```
riverar:
```suggestion
Write-Host "Restoring obsolete headers that are no longer in the SDK..."
```

jevansaks:
Thanks @riverar for the review! I added some emitter.settings.rsp to make sure that there are no effective changes even though we're now using the right headers (and those headers possibly have the wrong annotations 😝).

riverar (approved):

riverar (approved):

manodasanW (approved):

PR #2175, opened 2025-11-28
Add additional globalization constants

### LINKED ISSUE #2171: Missing definition for LOCALE_INVARIANT and others
The LOCALE_* macros are somewhat complex, so they are manually defined here:
generation/WinSDK/manual/Globalization.cs

This misses the other macros from the same header file:

LOCALE_CUSTOM_DEFAULT    
LOCALE_CUSTOM_UNSPECIFIED
LOCALE_CUSTOM_UI_DEFAULT
LOCALE_NEUTRAL
LOCALE_INVARIANT

riverar:
Fixes #2171


jevansaks:
Do we need to also update the ChangesSinceLastRelease.txt or do these not get picked up in that script?

PR #2185, opened 2025-12-31
Correct close function for `FwpmEngineOpen0` handle

DoctorKrolic:
Fixes: https://github.com/microsoft/win32metadata/issues/2172

PR #2192, opened 2026-01-08
Add domain-specific handle types for WFP enumerations

DoctorKrolic:
Follow up to https://github.com/microsoft/win32metadata/pull/2185

Unfortunatelly, all handle close functions also accept an engine handle, so C# won't be able to generate `SafeHandle` wrappers for them, but at least this will stop C# from projecting them as `SafeFileHandle`s, which is completely wrong


jevansaks:
@doctorkrolic did you try taking the locally-built Windows.Win32.winmd and using it with cswin32? What does it change about the cswin32 generated function signatures for the functions that use these types? 

And then @riverar  does the Rust projection support handle types with a CloseApi that has more than 1 parameter? And by "support" I mean "do a reasonable thing even though the projection probably can't auto-create them correctly".

DoctorKrolic:
> What does it change about the cswin32 generated function signatures for the functions that use these types?

Example previous friendly overload:
```cs
internal static unsafe uint FwpmConnectionCreateEnumHandle0(SafeHandle engineHandle, [Optional] winmdroot.NetworkManagement.WindowsFilteringPlatform.FWPM_CONNECTION_ENUM_TEMPLATE0? enumTemplate, out Microsoft.Win32.SafeHandles.SafeFileHandle enumHandle)
```

Friendly overload with changes from this PR:
```cs
internal static unsafe uint FwpmConnectionCreateEnumHandle0(SafeHandle engineHandle, [Optional] winmdroot.NetworkManagement.WindowsFilteringPlatform.FWPM_CONNECTION_ENUM_TEMPLATE0? enumTemplate, winmdroot.NetworkManagement.WindowsFilteringPlatform.FWPM_CONNECTION_ENUM_HANDLE* enumHandle)
```

Note, that the last parameter is no longer a `SafeFileHandle`, which is a good thing since `SafeFileHandle` will try to close handle with `CloseHandle` API upon disposal/destruction, which will lead to unexpected result. With new metadata CsWin32 didn't turn pointer into a managed ref for some reason, but this is an issue of CsWin32

jevansaks (approved):

Issue #2202, opened 2026-01-30
Annotate struct size fields in metadata


DoctorKrolic:
It is common in windows API to have a "sizeof" field in a struct (usually called `dwSize`) for versioning. I believe metadata can introduce a new `[SizeOf]` attribute or similar and annotate such fields with it. Generators using metadata can then account for that attribute and automatically assign right value to the annotated field. For instance, CsWin32 can do that in a friendly function overload

riverar:
Sadly no. (We get this suggestion a lot.)

Automatic initialization like this wouldn't work universally. For example, `Shell_NotifyIcon` accepts multiple versions of `NOTIFYICONDATA`, which the API distinguishes using the `cbSize` field. An automatic initialization scheme would need to telepathically know which version the developer intended to use.

Issue #2203, opened 2026-02-02
Suspicious duplicate structs that differ in architecture and pack only


DoctorKrolic:
I was ivestigating the underlying issue for https://github.com/microsoft/CsWin32/issues/1617#issuecomment-3833225127. It turned out metadata defines 2 nearly identical `RASDIALPARAMSW` structs which differs only in supported architecture and pack size:
```diff
.class public sequential ansi sealed beforefieldinit Windows.Win32.NetworkManagement.Rras.RASDIALPARAMSW
	extends [netstandard]System.ValueType
{
	.custom instance void [Windows.Win32.winmd]Windows.Win32.Foundation.Metadata.UnicodeAttribute::.ctor() = (
		01 00 00 00
	)
	.custom instance void [Windows.Win32.winmd]Windows.Win32.Foundation.Metadata.SupportedArchitectureAttribute::.ctor(valuetype [Windows.Win32.winmd]Windows.Win32.Foundation.Metadata.Architecture) = (
+		01 00 06 00 00 00 00 00 // X64 | Arm64
-		01 00 01 00 00 00 00 00 // X86
	)
+	.pack 4
+	.size 0

	// Fields
	.field public uint32 dwSize
	.field public char[0...256] szEntryName
	.field public char[0...128] szPhoneNumber
	.field public char[0...128] szCallbackNumber
	.field public char[0...256] szUserName
	.field public char[0...256] szPassword
	.field public char[0...15] szDomain
	.field public uint32 dwSubEntry
	.field public native uint dwCallbackId
	.field public uint32 dwIfIndex
	.field public valuetype [Windows.Win32.winmd]Windows.Win32.Foundation.PWSTR szEncPassword

} // end of class Windows.Win32.NetworkManagement.Rras.RASDIALPARAMSW
```
This seems suspicious. The struct has nothing to do with memory or any other hardware access, so it seems wrong that it is emitted as architecture-specific. I am not an expert in such low-level stuff as memory alignment on different architectures, but I have a suspicion that on x86 `.pack 4` is a default value, so it isn't emitted explicitly and that makes the wrong split later on. If true, both definitions can be collapsed into a single one without need for architecture-specific structs. I would like to either confirm or deny my suspicion

DoctorKrolic:
[...]

PR #2206, opened 2026-02-03
Improve heuristics for struct size field detection and annotate several structs by hand

DoctorKrolic:
Let's say this fixes https://github.com/microsoft/win32metadata/issues/2202


scripts/ChangesSinceLastRelease.txt
```diff
@@ -65,3 +65,685 @@ Windows.Win32.NetworkManagement.WindowsFilteringPlatform.IPSEC_SA_CONTEXT_ENUM_H
...
+Windows.Win32.NetworkManagement.Rras.PPP_IPV6_CP_INFO : [Documentation(https://learn.microsoft.com/windows/win32/api/mprapi/ns-mprapi-ppp_ipv6_cp_info)] => [Documentation(https://learn.microsoft.com/windows/win32/api/mprapi/ns-mprapi-ppp_ipv6_cp_info),StructSizeField(dwSize)]
+Windows.Win32.NetworkManagement.Rras.RAS_CONNECTION_3 : [Documentation(https://learn.microsoft.com/windows/win32/api/mprapi/ns-mprapi-ras_connection_3)] => [Documentation(https://learn.microsoft.com/windows/win32/api/mprapi/ns-mprapi-ras_connection_3),StructSizeField(dwSize)]
+Windows.Win32.NetworkManagement.Rras.RAS_STATS : [Documentation(https://learn.microsoft.com/windows/win32/api/ras/ns-ras-ras_stats)] => [Documentation(https://learn.microsoft.com/windows/win32/api/ras/ns-ras-ras_stats),StructSizeField(dwSize)]
+Windows.Win32.NetworkManagement.Rras.RASADPARAMS removed
+Windows.Win32.NetworkManagement.Rras.RASADPARAMS(X64, Arm64) added
+Windows.Win32.NetworkManagement.Rras.RASADPARAMS(X86) added
```
DoctorKrolic: This is a bug BTW. Adding `[StructSizeField]` attribute has nothing to do with CPU architecture and should not cause a split. This is at least somewhat related to https://github.com/microsoft/win32metadata/issues/2203

jevansaks: Hmmm yeah that's weird. I don't know how the tooling is determining there's a layout difference. I haven't had time to think about or look into solving the architecture layout discrepancies.

jevansaks:
This is a great improvement, thanks for taking this on. Before we merge this in we should figure out why the attribute is causing that struct to become architecture specific, since devs get upset when that happens.

scripts/ChangesSinceLastRelease.txt
DoctorKrolic: I debugged the project a little bit and found out the problem is here:
https://github.com/microsoft/win32metadata/blob/f06337c77e3fd3bdd3b5a5e90e0f720a8e00d90f/sources/ClangSharpSourceToWinmd/CrossArchSyntaxMap.cs#L324-L331
Since `RASADPARAMS` didn't have any attributes before this PR and we are adding `[StructSizeField]`, this branch is no longer taken for that struct. It appears to be the actual cause of #2203 as well, since `RASDIALPARAMSW` has a `[Unicode]` attribute which also breaks this code path. I am not sure why the logic is as it is though. I don't understand why any unrelated attribute makes the cross-arch merge unable to be performed. Can you please take a look?

PR #2170, opened 2025-11-26
Upgrade D3D12 Agility SDK to 1.618.5

MarijnS95:

PR #2224, opened 2026-02-26
Upgrade D3D12 Agility SDK to 1.619.1

MarijnS95:
Depends on https://github.com/microsoft/win32metadata/pull/2170

https://devblogs.microsoft.com/directx/shader-model-6-9-retail-and-more/


jevansaks (approved):

MarijnS95:
Thanks @jevansaks. This PR is purely based on top of #2170 so nothing is lost in the end. We just don't have a separate commit in history upgrading from `1.618.2` to `1.618.5` ahead of `1.619.1`, but that is unlikely to be useful to anyone.

PR #2284, opened 2026-08-10
ci: tolerate unavailable baseline artifacts

vineeththomasalex:
## What changed
- Treat an expired target-branch `winmd` artifact as unavailable.
- Catch baseline artifact download/extraction failures and continue with the existing NuGet baseline fallback.
- Increase the `winmd` artifact retention from 30 to the maximum 90 days.

## Why
The PR validation job failed after all build and test steps succeeded because the baseline artifact download step propagated an artifact-availability error. Baseline retrieval is auxiliary to API diff generation and should not fail the job when the NuGet fallback is available. Retaining the artifact for 90 days also reduces how often the fallback is needed.

## Validation
- `git diff --check` passes.
- The final PR diff contains only the intended fallback-handling and retention changes.
- Build, packaging, and tests were already successful in the referenced CI run before the baseline-download step failed.

This is a draft for review.

### DIFF
--- a/.github/workflows/pr-validation.yml
+++ b/.github/workflows/pr-validation.yml
@@ -180,24 +180,33 @@ jobs:
               return;
             }
 
+            if (winmdArtifact.expired) {
+              core.warning('The baseline winmd artifact has expired. Will fall back to NuGet baseline.');
+              return;
+            }
+
             // Download and extract
-            const download = await github.rest.actions.downloadArtifact({
-              owner: context.repo.owner,
-              repo: context.repo.repo,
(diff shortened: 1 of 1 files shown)


jevansaks:
﻿The fallback handling looks good. Could we also increase the `winmd` artifact retention from 30 days to the maximum 90 days in this PR?

```yaml
retention-days: 90
```

That should make the preferred main-branch baseline available through longer gaps between runs, while this PR's fallback still protects us after 90 days or from other artifact failures.

jevansaks (approved):

PR #2287, opened 2026-08-14
Upgrade winmd generation to ClangSharp 21

jevansaks:
## Summary
- upgrade the embedded scraper to ClangSharp 21.1.8.4 and baseline the entire WinmdGenerator SDK and repository toolchain on .NET 10
- support VS 2026/MSVC 14.51 headers and normalize ClangSharp 21 output without changing existing callback signatures, inheritance, anonymous-record identities, or metadata names
- restore ClangSharp response-file compatibility broken by the hosted-generator migration, including OSClient's `-std c++20`, short aliases, multiline values, and inline `--option=value` forms
- package one explicit `tools/net10.0` toolset plus the net10 Windows scraper/LLVM runtime; consumers no longer need both .NET 8 and .NET 10

## Parser and compatibility decisions
- language and standard options now reach both libclang and `PInvokeGeneratorConfiguration`; C++17 remains the default only when the consumer does not select a standard
- unsupported long response-file options fail explicitly instead of being silently discarded
- `Media.DShow` and `Mf` select C++20 locally because Clang 21 rejects a DXVA signed-shift integral constant expression under C++17; `dxva.h` and `dxva9typ.h` remain exactly as supplied by the SDK
- secondary COM bases represented by ClangSharp 21 as base-subobject fields participate in native-inheritance matching, preserving the existing projected hierarchy
- compatibility controls preserve cross-partition function-pointer levels, the correct `LIBID_SystemMonitor`, existing anonymous-record names, and an acyclic UMS projection

## WinMD comparison
A controlled Clang 17 and Clang 21 generation used the same source tree and VS 2026/MSVC headers. Both outputs contain 35,145 types. The only semantic changes are 19 `LIBID_*` constants recovered by ClangSharp 21; the apparent removal/addition of `LIBID_SystemMonitor` in the textual diff is ordering-only and retains its correct value.

The final single-.NET-10 build has a byte-for-byte identical 35,145-type declaration dump to the previously validated Clang 21 candidate. Candidate WinMD SHA-256: `32328C5DBBC450C03A4DEBF31D493895AE86D6E6B80E1048BF93E5CE0FD8F6C0`.

## .NET and packaging
ClangSharp 21 targets .NET 10, so WinmdGenerator now requires .NET 10 consistently rather than splitting its scraper and supporting tools across two runtimes. `global.json` selects SDK 10.0.303, all generator tools/tasks/tests target `net10.0`, SDK paths resolve under `tools/net10.0`, and CI installs only the SDK selected by `global.json`.

The generated package has 69 entries: zero `net8.0` entries, 38 explicit `net10.0` tool/runtime entries, and zero test artifacts.

## Validation
- clean `DoAll.ps1 -Clean -ExcludePackages -ExcludeSamples`
- MetadataUtils.Tests: 17 passed
- Win32MetadataScraperTests: 50 passed
- Windows.Win32.Tests: 13 passed
- packaged WinmdGenerator generated the DiaSdk sample successfully using only the .NET 10 tool layout
- final WinMD declaration dump is byte-for-byte identical to the validated Clang 21 candidate


jevansaks:
﻿## OSClient validation

Validated the packages from PR Validation run `33194187825` against `OSClientCsWin32Metadata` at OSClient `release/main` commit `1092104461841ff04f4b6e9fd571e306f9c2572a`.

- Package versions: `Microsoft.Windows.WinmdGenerator 0.71.26-preview-g38b8657306` and `Microsoft.Windows.SDK.Win32Metadata 71.0.26-preview-g38b8657306`
- VS2026: generation succeeded with VS 18.8 / MSVC 19.51 and .NET SDK 10.0.303; 0 warnings, 0 errors
- VS2022 toolchain: generation succeeded with the VS2022 17.14 v143 toolset (MSVC 19.44.35228) extracted from the official Build Tools layout; 0 errors
- The OSClient `libmappings.rsp` two-line `-std` / `c++20` setting was consumed by the corrected response-file parser
- Both outputs contain the same 52 sorted C# declarations. Binary hashes differ only because the emitter writes a per-build PE timestamp and MVID; no API difference exists.

A registered VS2022 instance could not be installed non-interactively because elevation was declined, so the VS2022 check used the official portable v143 compiler/header payload rather than a registered IDE. Separately, this PR's complete package/sample/test workflow passed on GitHub's `windows-2022` runner.

jonwis (approved):

PR #2316, opened 2026-09-16
Annotate Audio SDK headers

jevansaks:
## Summary

- preserve `IAudioClient::IsFormatSupported` result semantics and mark `GetService` as a COM out pointer
- shift Audio device, multimedia, and sound flags into guarded unscoped enums with direct native typing
- preserve the original `DWORD` width and signedness and keep loose constants associated with their enums

## Validation

- rebased directly onto FINAL/IMMUTABLE AppxPackaging `b33d844144d982cf5b2996fc49146f50574bcbc9`
- verified strict ancestry through partition 001 `c95957a1d2`, partition 002 `936d9c7217`, and collapsed partition 003/004 `0f1bb6d70988d6519cc9b9183138b003758c04df`
- isolated delta is exactly three Audio commits adding four consolidated patch files: `Audioclient.h`, `mmdeviceapi.h`, `mmeapi.h`, and `playsoundapi.h`
- legacy Audio supported-OS target audit: expected set `{}`, present set `{}`, 0 missing, 0 unexpected, 0 mismatched; all four patches contain zero `_Win32_SupportedOS_`, `SupportedOS`, or `MIDL_INTERFACE` entries
- `scripts\RecompileIdlFilesForScraping.ps1` applied all 18 stacked patches successfully (18 applied, 0 failed)
- all four Audio patches pass reverse check, reverse apply, forward check, and forward reapply
- `scripts\Generate-WindowsRsWinmd.ps1 -Partition Audio -Architecture x64,x86,arm64 -OutputWinmd .\bin\a.winmd` passes with 6,595 types, 4,314 functions, and 22,735 constants
- generated Audio headers parse with MSVC `/Zs` for x64, x86, and ARM64; ARM64 emits only the existing `__MACHINE` and `__MACHINEWVMPURE` redefinition warnings from `intrinfix.h`
- `cargo test --release --locked --manifest-path tools\rust\Cargo.toml` passes 15/15 tests
- working tree is clean; force-with-lease push completed and the local/remote full SHA matches
- Defender previously stopped the scraper executable before parsing with OS error 225; after the approved exclusion was enabled, executable ingestion completed successfully without bypassing Defender
- a short output stem (`a.winmd`) is used to avoid Windows MAX_PATH failures in generated LLVM resource-header paths


jevansaks:
Superseded by aggregate draft #2334, which preserves the exact immutable cumulative endpoint for partitions 001-010.

